Issue #27, opened 2018-06-13
roles without members does not shown in roles list


alex-vento:
Hello! In our case we have some roles without members and its does not shown in roles list, but if request an empty role detail page by its id, all works fine

sadysnaat:
I think this is because keto selects roles from keto_role_member table, and not from  keto_role table

https://github.com/ory/keto/blob/48dd9f1ffbeaa99ac8dc27085c5a50f9244bf9c3/role/manager_sql.go#L63-L70

https://github.com/ory/keto/blob/48dd9f1ffbeaa99ac8dc27085c5a50f9244bf9c3/role/manager_sql.go#L185-L191

arekkas:
Yeah, seems like a bug! Thanks for reporting and investigating!

PR #28, opened 2018-06-14
roles: List roles from keto_role table

sadysnaat:
Closes https://github.com/ory/keto/issues/27


arekkas:
Thanks! Could you also add a test case so this bug doesn't re-appear in the future? You could do that by creating a new role [after this line](https://github.com/ory/keto/blob/master/role/manager_test_helper.go#L81) and then check for the number of roles with `ds, err = m.ListRoles(100, 0)` which should be 2 (we are creating role 1, then role 2, then delete role 1, then create a new role without members and e.g. id 3 -> so 2 roles in total).

Thanks!

arekkas:

Issue #25, opened 2018-06-12
Wrong oauth introspect url


Detrous:
In help command instrospect url is “hydra-url/oauth2/introspection”, but introspect on hydra work at “hydra-url/oauth2/introspect”

“Example: AUTHENTICATOR_OAUTH2_INTROSPECTION_INTROSPECTION_URL=https://my-server/oauth2/introspection”

arekkas:
Thank you!

PR #39, opened 2018-08-01
cmd: Fixes typo in help command in env var name

RomanMinkin:
Related to #25 which has not been fixes completely. It should be `AUTHENTICATOR_OAUTH2_INTROSPECTION_URL` with one `INTROSPECTION` word not two.


arekkas:
Thanks!

Issue #47, opened 2018-09-04
Moving forward with ORY Keto


aeneasr:
I recently (re-)discovered the [OPA project](https://github.com/open-policy-agent/opa). This issue is about deprecating the ORY Ladon engine and aligning ORY Keto with OPA. The decision is not yet made and we are looking for valuable input regarding this.

OPA allows you to write authorization logic using a language specifically designed for that, called `rego`. Syntax is very go-like. Due to this, OPA is capable of providing all sorts of authorization mechanisms, like [RBAC, ABAC, ACL, AWS IAM Policies, and more](https://www.openpolicyagent.org/docs/comparison-to-other-systems.html). In fact, I believe that ORY Ladon's logic is implementable using `rego`. I'm not sure if that holds true for *conditions* which still needs verification from my side.

Let's take a look at the current downsides of each project.

I believe that policy documents as implemented by ORY Ladon are very powerful, but also very complicated. Many developers struggle with proper resource & action naming. I think that regular expressions have their place here, but many developers struggle with writing and testing regular expressions and variable substitution is very flaky (currently only used in the ORY Oathkeeper adapter for ORY Keto iirc) from a ux perspective. Also, regular expressions do not scale well, especially if read from the database. I think we can fix this with caching, but that is not fixing the problem itself, only the symptom.

On the same hand, OPA is limited. I think `rego` is great for developers that really want to jump into this (like me). But it's a new syntax, new language, and new tools. I think the language is not incredibly intuitive and not always readable:

I believe that policy documents are, in general, quite complicated. In my opinion, rego has a steep learning curve as well. Can you tell me *immediately* what this does?

```
sod_roles = [
    ["create-payment", "approve-payment"],
    ["create-vendor", "pay-vendor"]
]

sod_violation[user] {
    role1 = user_role[user][_]
    role2 = user_role[user][_]
    sod_roles[_] = [role1, role2]
}
```

At least I can not - the point I'm trying to make is, you have to learn this.

OPA comes with a REST API but it's really a parser and execution engine. It parses `rego` files and executes the logic based on data you provide. The result is always true or false, depending on the authorization result.

The server is limited. It stores everything in-memory so pushing logic to the server is not realistic with more than one server running. Instead, you'll probably have to write a CI pipeline which builds a docker image that has all your policy definitions. IMO that can be very nice, especially if you have `rego test` as part of that pipeline. But, from my experience, most people do not want or know how to do these things.

[...]

PR #48, opened 2018-09-23
Switch to rego as policy decision engine

aeneasr:
This is an experimental PR which aims to deprecate the ORY Ladon SDK in favor of  Rego, the language behind the [Open Policy Agent](http://openpolicyagent.org/).

This PR is the an attempt at improving the ORY Keto functionality and making it possible to add more access control mechanisms such as ACL, RBAC, ABAC, ... to the mix without writing a lot of code.

This PR will break some backwards compatibility regarding API design, but it will keep backwards compatibility with ORY Ladon's Access Control Policies!

* [x] Implement access control policies using rego
* [x] Implement adapters for rego
  * [x] Storage
    * [x] In-Memory
    * [x] SQL
  * [ ] *Reporters*
* [x] Implement API handlers for CRUD'ing data (e.g. policies, roles) and making requests like the warden

I think it makes sense to deprecate the authentication handlers as well. They have a lot of configuration overhead and might be added at a later stage as part of the rego logic. Please complain if you rely on them and need them in future releases.

This is related to #47


MOZGIII:
+1,909 −29,558 :+1:

fredbi:
I have not analyzed the pros and cons of this PR: I assume that, following previous discussions, there are good reasons for this move.

I just don't understand why such heavy rework has to come near the end of a series of beta releases, which should _normally_  converge toward a stable, rock-solid v1 for the whole ory stack.

Since this is still named _experimental_, I think this should be released in an alternate branch.

What do you think?

aeneasr:
Hey, there has been an extensive discussion (50 comments) about this here: https://github.com/ory/keto/issues/47

This PR is not about fancyness, it is about improving an otherwise limited product. The versioning is definitely an issue. I understand that putting something as version 1.0 should show confidence in the product by the maintainers. But we moved to this version not because we thought the product is finished, but because we wanted all products to work well together (Oathkeeper, Hydra, Keto). To ensure that and to avoid bad decisions with random versioning which have made e.g. Angular very hard to use.

In my experience as a maintainer of this technology stack I know that there will be a migration path for people already using Keto. This migration path will be a bit of work but the upsides you'll get will be tremendous. Making hard decisions is never easy, but if they're worth it they have to be made. The current version of Keto has serious issues wrt to database scalability and availability. In my opinion, it will not be easily possible to migrate to a new schema without having a migration path anyways. So why not take this opportunity and improve the product as a whole. I'm happy to help people using this in prod already to migrate.

The current plan for future releases of Keto and Oathkeeper, both of which are still evolving, is to tag them as `-preview` starting with the rc versioning. There will be a clear message in the history of each release what this means (e.g. future changes with migration paths) and we'll communicate that better.

You are, however, right as this should not be rushed. So maybe having this in a "next" branch is a good idea for now. But I really want people to be aware that this is what Keto will look like in the future, the old version just doesn't cut it. I'll take a bit of time to think about the best approach here, your input definitely helps and is welcomed. I hope we can find the best solution for moving this forward.

PR #50, opened 2018-10-11
cmd: TLS environment variables

fredbi:
* Makes TLS environment variables consistent with hydra and oathkeeper (see ory/oathkeeper#121)

Signed-off-by: Frederic BIDON <frederic@oneconcern.com>


aeneasr:
Thank you! However, I'm closing this in favor of #48 where the issue is addressed as well!

fredbi:
Ok so here is my 2 cents to this debate: do not break trust that is slowly built in this suite of tools.

Stuff such as #48 is exciting -- so I'd say let's release this as a different branch ("tip of the edge" or whatever).

Or even, as I suggested in your original "straw man" issue on this new replacer for ladon, a keto2 (or "cuato"?) that would be yet another standalone authorizer for oathkeeper.

Issue #121, opened 2019-07-13
GetOryAccessControlPolicyRoles by member id as well


sum2000:
**Is your feature request related to a problem? Please describe.**

More and more of our clients are asking for getting roles for a specific member or a group.

**Describe the solution you'd like**

Have an API `GetOryAccessControlPolicyRoles(flavor, Member ID/GroupID)` to get all the roles associated with a certain member

**Describe alternatives you've considered**

Our current workaround involves getting all the roles using `ListOryAccessControlPolicyRoles` and filter through specific subjects/members. This is not a secure way and also a memory intensive operation. Another potential way I thought was to read the rego_data directly which basically means having another keto API to support this.

**Additional context**

I have already started working on this so if you would like I can submit a PR for this. I am pretty much only waiting to see if you already have something in the pipeline or have any strong objection against having this API.

aeneasr:
Yup that makes sense, would you be up for a PR?

sum2000:
Yes, I will try to get it in! Thanks

monarkatfactly:
Hey @sum2000 any update on this issue?

sum2000:
It's already merged

monarkatfactly:
If I understand correctly using this API we can get all policies by subject?, because there is no mentions in doc about this feature.

malikvikrant:
Hey @sum2000 do you know how I can access this feature to get the roles based on member ID. I don't see anything related to this in document.

Issue #20, opened 2018-06-09
Policy filters


sebastian-schlecht:
Hey,

first of all thank you for this service, it seems awesome! Is there a way (planned) to filter policies, i.e. give me all the policies that match `some-scope:some-resource:<.*>`?

Thanks and cheers,
Sebastian

arekkas:
Not right now, would be a smart addition though!

sebastian-schlecht:
Alright, thanks!

arekkas:
You can keep this issue open until it's resolved :)

cobraz:
Hello there!

I'm working on a project where this is needed as well! :-)
Hola if you need help!

aeneasr:
Any ideas how to implement that? :)

sum2000:
@aeneasr this sounds similar to #121 , I can take it up once my PR #124 is approved. I am thinking to modify `ListOryAccessControlPolicies` to accept filters which will be `Subject`,` Resources`, `Actions`. Not sure if we want to add `Conditions` as well. On the `HTTP` side a query will look like `/engines/acp/ory/{flavor}/policies?subject=<some-subject>&resource=<some_resource>`. Although supporting advanced regex queries like `/engines/acp/ory/{flavor}/policies?subject=sum*&resource=<some_resource>` needs to be well thought out with possibly having a regex evaluator in storage. I am also a bit confused about the security aspect of it, how do we validate that requestor can get certain rows of the database or not.

aeneasr:
I think filters could be implemented here. The problem of course is that we either do a string against regex match, or treat regex as strings. I'm not sure if the second is really all that useful?

> I am also a bit confused about the security aspect of it, how do we validate that requestor can get certain rows of the database or not? My guess is API consumers have to validate user on their end.

Good point, right now everyone would be allowed to read all policies. Maybe you could use Keto to limit that by itself (inception!) - but yeah would need some thinking ;)

aeneasr:
I am closing this issue as our Google Zanzibar-based refactoring is scheduled to be released soon. Ory Keto up to version 0.5 will go in hibernation mode and receive only critical security patches.

PR #124, opened 2019-07-20
Roles and policies filter

sum2000:
## Related issue

#121 #20

## Proposed changes

Added support for filtering roles and policies as discussed with @aeneasr.

## Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.md)
- [x] I have read the [security policy](SECURITY.md)
- [x] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security
vulnerability, I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation within the code base (if appropriate)
- [x] I have documented my changes in the [developer guide](https://github.com/ory/docs) (if appropriate)

## Further comments

I messed up history rewrite in my original PR, resubmitting this one


storage/manager.go
```diff
@@ -12,7 +12,7 @@ import (
 
 type Manager interface {
 	Get(ctx context.Context, collection string, key string, value interface{}) error
-	List(ctx context.Context, collection string, value interface{}, limit, offset int) error
```
aeneasr: This `List` method is a generic. It can fetch policies, roles, and so on. Adding `member` here makes this interface function no longer generic but specific to the "list roles with members" method. Therefore, this change must be reversed.

Instead, what you could do, would be adding a `filter` property with type

```go
type Filters map[string]interface{}
```

and allowing filters to be passed here:

```go
List(ctx context.Context, collection string, value interface{}, filters Filters, limit, offset int) error
```

Then, in e.g. SQL you could construct a where clause using the filter, using the map key as the field name and the map value as the value for the field and making an "AND" clause if there are multiple filters.

Alternatively, you could even type the value part of the Filter, e.g.:

```go
type Filter struct {
  Value interface{}
  Foo string // whatever
}

type Filters map[string]Filter
```

Thank you!

engine/ladon/handler.go
```diff
@@ -432,11 +433,11 @@ func (e *Engine) policiesList(ctx context.Context, r *http.Request, ps httproute
 	if err != nil {
 		return nil, err
 	}
-
 	return &kstorage.ListRequest{
 		Collection: policyCollection(f),
 		Value:      &p,
 	}, nil
+
```
aeneasr: Remove line break

engine/ladon/handler.go
```diff
@@ -432,11 +433,11 @@ func (e *Engine) policiesList(ctx context.Context, r *http.Request, ps httproute
 	if err != nil {
 		return nil, err
 	}
-
```
aeneasr: Add line break

storage/manager.go
sum2000: Sure, that makes sense. Although, I do have some reservations on why are CRUD functions generic here. We lose type assertions by passing the value as `interface{}`. Implementing a filter which matches against JSON document can lead to unintended consequences during run time.

aeneasr: Yeah, you're right, I also overlooked how the query is constructed:

```
 		query = "SELECT document FROM rego_data WHERE collection=? AND (length(?)=0 OR json_extract(document,'$') LIKE ?) ORDER BY id ASC LIMIT ? OFFSET ?"
 	case dbal.DriverPostgreSQL:
 		query = "SELECT document FROM rego_data WHERE collection=? AND (length(?)=0 OR document::text LIKE ?) ORDER BY id ASC LIMIT ? OFFSET ?"
```

Since the database doesn't use any index here it will be as fast as loading these to memory and filtering them in Go. If we get performance issues there's always caching. Long-term we will replace this storage adapter anyways!

So how about doing this instead:

```go
type ListRequest struct {
	Collection string
	Value      interface{}
+.     FilterFunc(*ListRequest)
}

+ func (*l ListRequest) ListRequest Filter() *ListRequest {
+    if l.FilterFunc != nil {
+       l.FilterFunc(l)
+   }
+   return l
+ }


func (h *Handler) List(factory func(context.Context, *http.Request, httprouter.Params) (*ListRequest, error)) httprouter.Handle {
	return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
		ctx := r.Context()
		l, err := factory(ctx, r, ps)
		if err != nil {
			h.h.WriteError(w, r, err)
			return
		}

		limit, offset := pagination.Parse(r, 100, 0, 500)
		if err := h.s.List(ctx, l.Collection, l.Value, limit, offset); err != nil {
			h.h.WriteError(w, r, err)
			return
		}

-		h.h.Write(w, r, l.Value)
+		h.h.Write(w, r, l.Filter().Value)
	}
}
```

And use the Filter function to sort for roles by members? We would need to do a type assertion here but it's safe to say that the type will always validate to true, and if not it's a programming error (should thus panic).

sum2000: I like this approach more than the previous. I also like the approach of just indexing the JSON data using the approach mentioned [here](https://docs.microsoft.com/en-us/sql/relational-databases/json/index-json-data?view=sql-server-2017). Regarding the type-assertion; the problem is not type-assertion after `List()` returns, it is before `Upsert()` is called as I can pass `value interface{} `as literally anything, but in reality, this `value interface{} `should point to `type Role`. In the interest of time, I can move forward with the filter approach.

aeneasr:
Let me know when this is good for 👁  :)

sum2000:
Swamped at work, will probably finish this tomorrow.

engine/ladon/doc.go
```diff
@@ -245,6 +245,11 @@ type listOryAccessControlPolicyRoles struct {
 	// required: true
 	Flavor string `json:"flavor"`
 
+	// The Member (ID) for which the roles are to be listed (Optional).
```
sum2000: I tried adding Member []string in Roles and Sub, Action, Res in Policies but swagger generate won't add it in the spec. Can you share your process for adding URL query params in the API in the swagger doc

storage/handler.go
```diff
@@ -87,15 +87,21 @@ func ListByQuery(l *ListRequest, m map[string][]string) {
 	case *Roles:
 		var res Roles
 		for _, role := range *val {
-			res = append(res, *role.withMembers(m["member"]).withIDs(m["id"]))
+			filteredRole := role.withMembers(m["members"]).withIDs(m["id"])
```
sum2000: A note on the filter: If someone passed unknown url query param to roles or policies list API then that isn't handled here. All values will be returned unless there is some parameter validation happening before filter is called.

sum2000: Also, not sure if query params should be singular or plural. I added support for plural i.e. "members" not "member"

sum2000:
Good to review apart from the comments I mentioned

sum2000:
@aeneasr

aeneasr:
Sorry I completely dropped the ball on this one! Reviewing now

engine/ladon/doc.go
aeneasr: Since this is `in: query` it can only be a raw string _I think_ (not sure though).

storage/policy.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Policies is an array of policies.
+//
+// swagger:ignore
+type Policies []Policy
+
+// Policy specifies an ORY Access Policy document.
+//
+// swagger:ignore
+type Policy struct {
```
aeneasr: Since these are 1:1 copies of https://github.com/ory/keto/blob/master/engine/ladon/policy.go#L31 I think it would make sense to move these completely (delete the files ladon/policy etc).

engine/ladon/doc.go
sum2000: Even that doesn't work

aeneasr: Hm, maybe I misunderstood the problem? Are you trying for this to become a `[]string`?

sum2000: The problem is that swagger spec generation is not adding any new field I add to these request structs. I tried both Member string and Member []string but after running make sdk the swagger query params for list only has offset and limit.

storage/handler.go
sum2000: can you clarify this? @aeneasr

storage/policy.go
sum2000: Regarding tests, I cant complete those tests unless swagger spec is correctly generated and the new sdk has the updated fields. So could you help me resolve that issue at the earliest?

engine/ladon/doc.go
aeneasr: Ok I will check it out now

docs/api.swagger.json
```diff
@@ -287,6 +287,13 @@
             "in": "path",
             "required": true
           },
+          {
+            "type": "string",
```
aeneasr: @sum2000 the member parameter is added properly here

sdk/go/keto/client/engines/list_ory_access_control_policy_roles_parameters.go
```diff
@@ -138,6 +143,17 @@ func (o *ListOryAccessControlPolicyRolesParams) SetLimit(limit *int64) {
 	o.Limit = limit
 }
 
+// WithMember adds the member to the list ory access control policy roles params
+func (o *ListOryAccessControlPolicyRolesParams) WithMember(member *string) *ListOryAccessControlPolicyRolesParams {
```
aeneasr: @sum2000 this is also added here?

aeneasr:
I'm not sure what the problem is - maybe I misunderstood it - but it seems that the `member` parameter is propagated properly in the SDK? Maybe you could lay out the problem again in one comment, I think I must have missed something.

storage/handler.go
aeneasr: Let's keep it singular. Plural can be both AND and OR and we don't want to deal with that right now.

Regarding unknown parameters - that's fine. If it's unknown, it will be ignored.

sum2000: Okay, those were my concerns as well but we dont allow passing multiple values for let's say actions then the problem of calling list api multiple times is still there. The way I have written it if someone calls `../policies/list?actions="delete"&action="create"&subject="122"`, the filter will still work.

docs/api.swagger.json
sum2000: @aeneasr I added these manually to test the api but automatic generation from swagger spec generate doesn't add it on my system. If you notice, the policy queries are not added. I will uninstall go-swagger and reinstall it and try again.

sdk/go/keto/client/engines/list_ory_access_control_policy_roles_parameters.go
sum2000: see my commment above

docs/api.swagger.json
sum2000: Also, swagger 2.0 allows arrays of native types to be passed in query, so Member []string should work.

storage/handler.go
sum2000: @aeneasr

engine/ladon/doc.go
sum2000: Okay, finally I was able to solve that issue. The brew installed version of go-swagger had some caching issue and wouldn't generate new swagger spec on changes. I had to install go-swagger from source and then it worked as expected. Phewww!!!

sum2000:
@aeneasr ready for review again. A note on the filter method, if someone passing regex query in the filter params then they are treated as exact. This can be changed in future PR. Let's try to close this soon.

aeneasr:
Awesome - thank you for your patience and hard work, you rock! :)

sum2000:
@aeneasr My pleasure, many more PRs to come! :)

PR #151, opened 2020-01-23
Disable access logging for health endpoints

### LINKED ISSUE #150: Disable access logging or change level of logs for the health endpoints
I would like to be able to disable access logging for the health endpoints, so that I can make sense of my application logs again. Inside Kubernetes, the health endpoints of Keto are checked regularly, with each checks equal to two lines in its logfile.

I would propose either of these two solutions:

1. Disable access logging for the health endpoints, akin to [what Hydra already has implemented](https://github.com/ory/hydra/blob/fb31d017fcec2ee52e94c0e1b88a66e58ccc893b/docs/config.yaml#L137-L139).
2. Shift the level of severity for the logs from `info` to `debug` to avoid logging them in a "production" context.

As an example, here are two lines in my logs that I encounter on a regular basis (i.e. every couple of seconds):

```
{"msg": "started handling request", "request": "/health/ready", "method": "GET", "level": "info", "time": "2020-01-22T11:51:21Z", "remote": "127.0.0.1:50224"}
{"msg": "completed handling request", "text_status": "OK", "request": "/health/alive", "took": 78385,  "measure#keto": {"latency": 78385}, "method": "GET", "level": "info", "time": "2020-01-22T11:52:42Z", "remote": "127.0.0.1:37770", "status": 200}
```
_Note: Not the same event._

Thanks!

moritzheiber:
Fixes #150 

@aeneasr

### DIFF
--- a/cmd/server/serve.go
+++ b/cmd/server/serve.go
@@ -29,7 +29,6 @@ import (
 	"github.com/ory/x/logrusx"
 
 	"github.com/julienschmidt/httprouter"
-	negronilogrus "github.com/meatballhat/negroni-logrus"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


aeneasr:
Thank you!

PR #157, opened 2020-03-06
fix(driver):Update ory/x version

### LINKED ISSUE #145: Cannot connect to MySQL
**Describe the bug**

It is not possible to connect to MySQL using the DSN format.

**Reproducing the bug**

The issue is related to a change in Golang and was first reported here: https://github.com/golang-migrate/migrate/issues/264

**Expected behavior**

Keto should be able to connect to MySQL.

**Environment**

* Keto versions: Higher than v0.3.3 (it works in v0.3.3)
* Environment: Tested in Docker

**Additional context**

The issue has already been fixed: https://github.com/golang-migrate/migrate/pull/265
Most probably the solution is simply to update the Go version.

yuan1238y:
## Related issue
Fixes #145 again

## Proposed changes

Run `go get -u github.com/ory/x/sqlcon`

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)

`go test --short ./...`

image


aeneasr:
Ah yeah, this currently fails because we changed the driver in ory/x from postgres to pgx. I can take a look at this in about 10 days unless you're up for fixing this. I think all you need to do is replace `lib/pq` imports with `github.com/jackc/pgx/stdlib`

yuan1238y:
> Ah yeah, this currently fails because we changed the driver in ory/x from postgres to pgx. I can take a look at this in about 10 days unless you're up for fixing this. I think all you need to do is replace `lib/pq` imports with `github.com/jackc/pgx/stdlib`

I try to replace `lib/pq` but the test fails still.

aeneasr:
Ok, I am currently out of office and can not check out the changes locally but will be able to do so next week.

PR #250, opened 2020-10-01
fix: Use make() to initialise array of Policy and Role instead of leaving it uninitalised

### LINKED ISSUE #217: Policies and roles list should return empty array if list is empty
**Describe the bug**

Policies and roles list are returning `null` instead of empty list if all elements are filtered out.
This breaks api contract and generates an error in php client (for example)

**Reproducing the bug**

1. Send a GET request to `/engines/acp/ory/glob/policies` on a newly created keto server
2. Response is `null` instead of `[]`

xlanor:
Fixes #217 
This is an ongoing issue in `encoding/json`, whereby if an uninitialised array (such as policies, which is declared as `var p Policies`, or roles, which is declared as `var p Roles`) is attempted to be encoded using `json.Marshal`, it results in a `nil` value being returned instead of `[]`.

Given that json doesn't appear to be merging the currently open PR to add an option to return nil as [] or {} in when json.Marshal is used anytime soon, this patch should fix that.

I've updated the tests in filter_test.go and filter_helper_test.go to replace the expected nil values with the relevant empty Role/Policy object.

## Related issue

#217
https://github.com/golang/go/issues/27589
## Checklist

      the changes.
      works.

## Further comments

Below is the tracing process that I posted on #217 

I've been tracing this bug because it also affects us at work, and I've traced it to this
`func (h *Handler) List(factory func(context.Context, *http.Request, httprouter.Params) (*ListRequest, error)) httprouter.Handle` in `storage/handler.go` returns a `h.h.Write(w, r, l.Filter(m).Value)`, and in this case, `l.Filter(m).Value` returns `*storage.Policies`, which is declared as a type of []Policy .

`h`, in this case, which is a type of `herodot.Writer`, an interface, in this particular function, is a `herodot.JSONWriter`, and   `Write` in this case calls `WriteCode` [here](https://github.com/ory/herodot/blob/master/json.go). This tells me that there's an issue with `encoding/json`'s  `json.Marshal(e)`, where `e` refers to `l.Filter(m).Value`.

I believe this is due to https://github.com/golang/go/issues/27589, and a proposal has been made to modify `encoding/json` that is still under going review [here](https://go-review.googlesource.com/c/go/+/136761/) and awaiting a PR at https://github.com/golang/go/issues/37711.

At this point, I'm wondering if I should submit a PR to `ory/herodot`, to do the checking in `func WriteCode`. If it is a shared library, do you think that this will affect other projects that ory has?

There's a quick fix, as can be seen from https://github.com/CosmWasm/go-cosmwasm/pull/75, to quickly check the value, and if it's nil, do not even proceed with `JSON.Marshal`. This might actually be the fastest way to fix this, but I'm unsure about the potential impact, given that herodot is a shared repository.

Ultimately, the issue lies in `func ListByQuery` in `storage/handler.go`, which declares var res Roles and var res Policies without explicitly initialising them, as well as `func rolesList` and `func policiesList` in `engine/ladon/handler.go`, which does the same.

### DIFF
--- a/engine/ladon/handler.go
+++ b/engine/ladon/handler.go
@@ -290,8 +290,7 @@ func (e *Engine) Register(r *httprouter.Router) {
 }
 
 func (e *Engine) rolesList(ctx context.Context, r *http.Request, ps httprouter.Params) (*kstorage.ListRequest, error) {
-	var p kstorage.Roles
-
+	p := make([]kstorage.Role, 0)
 	f, err := flavor(ps)
 	if err != nil {
 		return nil, err
@@ -428,8 +427,8 @@ func (e *Engine) policiesCreate(ctx context.Context, r *http.Request, ps httprou
 }
 
 func (e *Engine) policiesList(ctx context.Context, r *http.Request, ps httprouter.Params) (*kstorage.ListRequest, error) {
-	var p kstorage.Policies
 
+	p := make([]kstorage.Policy, 0)
 	f, err := flavor(ps)
 	if err != nil {
 		return nil, err
--- a/engine/ladon/handler.go
+++ b/engine/ladon/handler.go
@@ -290,7 +290,7 @@ func (e *Engine) Register(r *httprouter.Router) {
 }
 
 func (e *Engine) rolesList(ctx context.Context, r *http.Request, ps httprouter.Params) (*kstorage.ListRequest, error) {
-	p := make([]kstorage.Role, 0)
+	p := make(kstorage.Roles, 0)
 	f, err := flavor(ps)
 	if err != nil {
 		return nil, err
@@ -428,7 +428,7 @@ func (e *Engine) policiesCreate(ctx context.Context, r *http.Request, ps httprou
 
 func (e *Engine) policiesList(ctx context.Context, r *http.Request, ps httprouter.Params) (*kstorage.ListRequest, error) {
 
-	p := make([]kstorage.Policy, 0)
+	p := make(kstorage.Policies, 0)
 	f, err := flavor(ps)
 	if err != nil {
 		return nil, err
--- a/storage/filter_helper_test.go
+++ b/storage/filter_helper_test.go
@@ -52,7 +52,7 @@ var (
 				Members: []string{"mem1", "mem2"},
 			},
 		},
-		nil,
+		Roles{},
 		rolReq,
 		rolReq,
 		rolReq,
@@ -93,7 +93,7 @@ var (
 				Resources: []string{"res1", "res2"},
 			},
 		},
-		nil,
-		nil,
+		Policies{},
+		Policies{},
 	}
 )
--- a/storage/filter_helper_test.go
+++ b/storage/filter_helper_test.go
@@ -35,7 +35,7 @@ var (
 		{"action": {"create"}, "subject": {"mem3"}, "resource": {"res3"}},
 		{"action": {"delete"}},
 	}
-	rolRes = []Roles{
+	rolRes = [][]Role{
 		[]Role{
 			{
 				ID:      "role1",
@@ -59,7 +59,7 @@ var (
 		rolReq,
 		rolReq,
 	}
-	polRes = []Policies{
+	polRes = [][]Policy{
 		polReq,
 		polReq,
 		polReq,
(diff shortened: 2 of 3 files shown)


engine/ladon/handler.go
```diff
@@ -290,8 +290,7 @@ func (e *Engine) Register(r *httprouter.Router) {
 }
 
 func (e *Engine) rolesList(ctx context.Context, r *http.Request, ps httprouter.Params) (*kstorage.ListRequest, error) {
-	var p kstorage.Roles
-
+	p := make([]kstorage.Role, 0)
```
aeneasr, line 293: It is possible to make types that have an underlying type of a slice or map:

```suggestion
	p := make(kstorage.Roles, 0)
```

engine/ladon/handler.go
```diff
@@ -428,8 +427,8 @@ func (e *Engine) policiesCreate(ctx context.Context, r *http.Request, ps httprou
 }
 
 func (e *Engine) policiesList(ctx context.Context, r *http.Request, ps httprouter.Params) (*kstorage.ListRequest, error) {
-	var p kstorage.Policies
 
+	p := make([]kstorage.Policy, 0)
```
aeneasr, line 431:
```suggestion
	p := make(kstorage.Policies, 0)
```

storage/filter_helper_test.go
```diff
@@ -1,7 +1,7 @@
 package storage
```
aeneasr, line 1: Revert the changes in this file :)

storage/handler.go
```diff
@@ -84,17 +84,17 @@ func (l *ListRequest) Filter(m map[string][]string) *ListRequest {
 
 func ListByQuery(l *ListRequest, m map[string][]string) {
 	switch val := l.Value.(type) {
-	case *Roles:
-		var res Roles
+	case *[]Role:
```
aeneasr, line 87:
```suggestion
	case *Roles:
```

storage/handler.go
```diff
@@ -84,17 +84,17 @@ func (l *ListRequest) Filter(m map[string][]string) *ListRequest {
...
 		for _, role := range *val {
 			filteredRole := role.withMembers(m["member"]).withIDs(m["id"])
 			if filteredRole != nil {
 				res = append(res, *filteredRole)
 			}
 		}
 		l.Value = &res
-	case *Policies:
-		var res Policies
+	case *[]Policy:
```
aeneasr, line 96:
```suggestion
	case *Policies:
```

storage/handler.go
```diff
@@ -84,17 +84,17 @@ func (l *ListRequest) Filter(m map[string][]string) *ListRequest {
 
 func ListByQuery(l *ListRequest, m map[string][]string) {
 	switch val := l.Value.(type) {
-	case *Roles:
-		var res Roles
+	case *[]Role:
+		res := make([]Role, 0)
```
aeneasr, line 88:
```suggestion
		res := make(Roles, 0)
```

storage/handler.go
```diff
@@ -84,17 +84,17 @@ func (l *ListRequest) Filter(m map[string][]string) *ListRequest {
...
 			filteredRole := role.withMembers(m["member"]).withIDs(m["id"])
 			if filteredRole != nil {
 				res = append(res, *filteredRole)
 			}
 		}
 		l.Value = &res
-	case *Policies:
-		var res Policies
+	case *[]Policy:
+		res := make([]Policy, 0)
```
aeneasr, line 97:
```suggestion
		res := make(Policies, 0)
```

aeneasr (commented):
Awesome, thank you for your contribution! This looks pretty good but I have some ideas how to improve it further :)

xlanor:
Ah, I didn't know that! It's late here now, I will push the changes that you've suggested tomorrow

storage/filter_helper_test.go
xlanor: Apart from reverting this, I have also changed the expected nil values to match the empty policies/roles that is returned now.

xlanor:
@aeneasr , The requested changes have been made!

aeneasr (approved):
Awesome 🎉 

PR #266, opened 2020-10-08
feat: implement basic zanzibar API contract

zepatrik:
This PR is an umbrella PR that keeps track of everything that goes into implementing a zanzibar-like next-gen Keto. It shows the current state of the zanzibar branch where the development happens. If you have general feedback/usecases/.... please let us know either in this PR or by opening an issue.

### DIFF
--- a/.docker/Dockerfile-sqlite
+++ b/.docker/Dockerfile-sqlite
@@ -1,12 +1,26 @@
 FROM alpine:3.13
 
+# Because this image is built for SQLite, we create /home/ory and /home/ory/sqlite which is owned by the ory user
+# and declare /home/ory/sqlite a volume.
+#
+# To get SQLite and Docker Volumes working with this image, mount the volume where SQLite should be written to at:
+#
+#   /home/ory/sqlite/some-file.
+
 RUN addgroup -S ory; \
     adduser -S ory -G ory -D  -h /home/ory -s /bin/nologin; \
     chown -R ory:ory /home/ory
-
 RUN apk add -U --no-cache ca-certificates
 
+WORKDIR /home/ory
+
 COPY keto /usr/bin/keto
+
+# By creating the sqlite folder as the ory user, the mounted volume will be owned by ory:ory, which
+# is required for read/write of SQLite.
+RUN mkdir -p /var/lib/sqlite
+RUN chown ory:ory /var/lib/sqlite
+VOLUME /var/lib/sqlite
 
 # Exposing the ory home directory to simplify passing in keto configuration (e.g. if the file $HOME/.keto.yaml
 # exists, it will be automatically used as the configuration file).
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -87,7 +87,6 @@ builds:
       - CGO_ENABLED=0
     goarch:
       - amd64
-      - 386
       - arm
       - arm64
     goarm:
@@ -95,7 +94,6 @@ builds:
       - 6
       - 7
     goos:
-      - freebsd
       - linux
       - windows
       - darwin


robinbraemer:
I was just about to build exactly this authorization system based off of Google’s Zanzibar paper, but then found the very good looking “zanzibar/expand-api” branches in the Keto project. 👍 

I would love to help and contribute to Keto’s ACL as I’m also familiar with the Zanzibar paper and began some thoughts about such system myself for my organization.

@zepatrik I assume the Zanzibar based ACL system will be a public API as well as provide the under layer for the planned RBAC, ABAC etc. systems as its that flexible enough?

Let me know if there is something I could help with in terms of implementing or as another code reviewer giving ideas and giving feedback!

zepatrik:
Awesome, do you want to join our slack for discussions? https://slack.ory.sh/
I would also be up to jump on a call at some point. Just PM me (Patrik (ory)) on slack.

For transparency, here are some answers to your questions:
1. There will be a public API that allows requests such as "is `user` allowed to do `relation` on `object`" (currently the check package).
1. There will be a public API to expand user sets (the expand package).
1. The ACL language itself is flexible enough to allow implementation of RBAC, ABAC, ... We want to evaluate it today to figure out the limitations now
1. For the database interface we will probably go with https://github.com/gobuffalo/pop as we had good experience with that. It allows us to support multiple databases at once, including CRDB or Yugabyte through the postgres driver. This way you can use them to get a cloudspanner like database.
1. The other optimizations you mentioned are planed for a later stage when we have the functionality working.
1. All the distributed and fan-out part of zanzibar will be added later as well. We try to use algorithms now that allow us to do the recursion over multiple nodes later on.
1. The whole idea of a zookie is not yet implemented as it heavily relies on the database. Naming in general is something we have to rework anyway.

robinbraemer (approved):

internal/persistence/sql/relationtuples.go
```diff
@@ -0,0 +1,157 @@
...
+	nextPageToken := pagination.encodeNextPageToken()
+	if len(internalRes) == 0 {
+		nextPageToken = pageTokenEnd
+	}
+	return internalRes, nextPageToken, nil
+}
+
+func (p *Persister) WriteRelationTuples(ctx context.Context, rs ...*relationtuple.InternalRelationTuple) error {
+	return p.transaction(ctx, func(ctx context.Context, tx *pop.Connection) error {
+		for _, r := range rs {
```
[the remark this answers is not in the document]
zepatrik: Sure, the sql persister will need some performance tests and improvements. Let's rework all of that in one go later on. It is not as easy unfortunately, as we support multiple databases there and need to evaluate the solution on all of them. Quick search for sqlite (the usual exception): https://stackoverflow.com/a/5009740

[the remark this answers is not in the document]
zepatrik: tracked as #396

api/keto/acl/v1alpha1/read_service.proto
```diff
@@ -61,7 +61,7 @@ message ListRelationTuplesRequest {
   string snaptoken = 3;
   // Optional. The maximum number of
   // RelationTuples to return in the response.
-  int32 page_size = 4;
+  uint32 page_size = 4;
```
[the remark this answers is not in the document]
zepatrik: this was fixed in #395 already, think you had a stale view in the review there :wink:

jon-whit:
@zepatrik I'd love to pitch in and work on this with you. I too am working on a similar design after having read the Zanzibar paper. I'm also trying to get this type of architecture built and scaled up to work for an internal use case at my company. So I'd have plenty of time to work on it.

I'd love to join the slack and collaborate with you weekly or bi-weekly.

aeneasr:
I would like to get this merged - what's the status on this?

zepatrik:
I updated the milestone: https://github.com/ory/keto/milestone/4
All of the issues are not too much work, just the last details for the release. I think we are at a point where we could rename the current master to e.g. `legacy` and make this the new master branch, what do you think?

aeneasr:
Yeah, I think that makes sense - setting up a legacy branch based on current master and then force-pushing the zanzibar branch.

docs/docs/secure.md
```diff
@@ -3,13 +3,13 @@ id: secure
...
 access control on their own. Any request made to any Keto API is considered
 authenticated, authorized, and is thus being executed. However, these endpoints
 are very sensitive as they define who is allowed to do what in your system.
 
 Please protect these endpoints using
-[ORY Oathkeeper](https://github.com/ory/oathkeeper) or a comparable API Gateway.
+[Ory Oathkeeper](https://github.com/ory/oathkeeper) or a comparable API Gateway.
 How you protect them, is up to you.
 
 If you require support for this, consider [asking us](mailto:hi@ory.sh).
```
aeneasr, line 15: Would probably link to GitHub discussions here instead

docs/sidebar.json
```diff
@@ -1,18 +1,23 @@
...
+    "secure",
+    "guides/simple-access-check-guide"
   ],
-  "Access Control Engines": [
-    "engines/index",
-    "engines/acl",
-    "engines/rbac",
-    "engines/acp-ory",
-    "engines/acp-aws"
+  "Concepts": [
```
aeneasr, line 11: Move concepts above guides

aeneasr (changes requested):
- [ ] Update the introduction page https://github.com/ory/keto/blob/master/docs/docs/index.md

docs/docs/examples/cat-videos.mdx
```diff
@@ -0,0 +1,156 @@
...
+First off, we get a request by an anonymous user that would like to view
+`/cats/2.mp4`. The client now has to ask Keto if that operation should be
+allowed or denied.
+
+```shell
+# Is "*" allowed to "view" the object "videos":"/cats/2.mp4"?
+keto check "*" view videos /cats/2.mp4
+# output:
+
+# false
```
aeneasr, line 96:
```suggestion
$ keto check "*" view videos /cats/2.mp4

Denied
```

contrib/cat-videos-example/up.sh
```diff
@@ -0,0 +1,26 @@
...
+
+Created all relation tuples. Now you can use the Keto CLI client like so:
+
+export KETO_GRPC_URL=\"127.0.0.1:4467\"
+keto relation-tuple get videos
+keto check \"*\" view videos /cats/1.mp4
+keto expand view videos /cats/2.mp4
+"
+
+sleep 10d
```
aeneasr, line 26: This kills the process after a couple of seconds on macOS. I would instead run the echo commands as a subroutine / background task with the relation-tuple create in front.

Makefile
```diff
@@ -1,65 +1,84 @@
...
+test-e2e:
+		go test -tags sqlite -failfast -v ./internal/e2e
+
+.PHONY: test-docs-samples
+test-docs-samples:
+		cd ./contrib/docs-code-samples \
+		&& \
+		npm i \
+		&& \
+		npm test
```
aeneasr, line 84: Please add a "make docker" target which builds the Dockerfile including SQLite!

Dockerfile
```diff
@@ -1,16 +1,43 @@
-# To compile this image manually run:
-#
-# $ packr; GO111MODULE=on GOOS=linux GOARCH=amd64 go build; docker build -t oryd/keto:latest .; rm keto; packr clean
-FROM alpine:3.9
+FROM golang:1.16-alpine AS builder
```
aeneasr, line 1: Please follow the structure of https://github.com/ory/kratos/tree/master/.docker

The main Dockerfile is an empty container where goreleaser injects its binary into.

docs/docs/examples/cat-videos.mdx
```diff
@@ -0,0 +1,156 @@
...
+$ keto help
+```
+
+Alternatively, using Docker automatically gets the required images:
+
+```shell
+# clone the repository if you don't have it yet
+git clone git@github.com:ory/keto.git && cd keto
+
+docker-compose -f contrib/cat-videos-example/docker-compose.yml up
```
aeneasr, line 38: This fails with several error messages, including the fallback URL.

docs/sidebar.json
```diff
@@ -1,18 +1,23 @@
...
+    "concepts/relation-tuples",
+    "concepts/namespaces",
+    "concepts/objects",
+    "concepts/subjects",
+    "concepts/graph-of-relations"
+  ],
+  "Examples": [
+    "examples/olymp-file-sharing",
+    "examples/cat-videos"
```
aeneasr, line 20: I think this should be the  quickstart!

aeneasr (changes requested):
Bildschirmfoto 2021-03-12 um 08 25 31

docs/docs/concepts/relation-tuples.mdx
```diff
@@ -0,0 +1,40 @@
...
+---
+
+import RelationTuplePrism from '@theme/RelationTuplePrism'
+RelationTuplePrism()
+
+Relation tuples are the underlying datatype of Ory Keto's access control
+language. They encode relations between [objects](./objects) and
+[subjects](./subjects). A relation tuple is associated with a
+[namespace](./namespaces) where its relation has to be defined and configured.
+The following BCNF grammar describes the encoding used in the documentation and
```
aeneasr, line 12: What is BCNF 😅 Maybe link to something?

docs/docs/concepts/relation-tuples.mdx
```diff
@@ -0,0 +1,40 @@
...
+within Ory Keto. Note that the namespace is often omitted in examples for
+readability, but is technically always required.
+
```
aeneasr, line 15:
```suggestion
within Ory Keto.

:::note

The namespace is often omitted in examples for
readability, but is technically always required.

:::

```

docs/docs/concepts/namespaces.mdx
```diff
@@ -0,0 +1,51 @@
...
+Vice versa, all relation tuples containing an object have to be in the same
+namespace to reference the same object.
+
+## Migrations
+
+Because namespaces each come with an individual configuration that can even
+modify some storage specific options, it is required to manually review and run
+migrations on namespace configuration updates. Please refer to the
+[namespace migration CLI reference](/TODO) and
+[running in production guide](/TODO) to learn more about that process.
```
aeneasr, line 51: Update links, write migration guide.

docs/docs/concepts/namespaces.mdx
```diff
@@ -0,0 +1,51 @@
+---
+title: Namespaces
+---
+
+import RelationTuplePrism from '@theme/RelationTuplePrism'
+RelationTuplePrism()
+
+Ory Keto knows the concept of namespaces to organize
+[relation tuples](./relation-tuples). Namespaces have a configuration that
+defines the relations, and some other important values ([see reference](/TODO)).
```
aeneasr, line 10: Update links

docs/docs/concepts/namespaces.mdx
```diff
@@ -0,0 +1,51 @@
...
+Unlike other applications, Ory Keto does **not** isolate namespaces. Especially,
+[subject sets](./subjects) can cross-reference from one namespace to another.
+The namespaces' purpose is to split up the data into coherent partitions, each
+with its corresponding configuration. Internally each namespace has its own
+table in the database to allow setting individual
+[storage specific options](/TODO).
```
aeneasr, line 16: Update links - basically just search for `/TODO`

aeneasr (commented):
Please replace all `/TODO` links and/or write the content for those links.

docs/docs/examples/olymp-file-sharing.mdx
```diff
@@ -0,0 +1,72 @@
...
+#### A basic, down-to-earth full feature example
+
+Consider a file sharing application called "Olymp Library". Each file is stored
+in a key-value store, where the key is a UUIDv4 (pseudorandom unique
+identifier), while the value is the metadata and content. The application uses
+Ory Keto to keep track of ownership and granted access on a per file level.
+
+:::note
+
+This example assumes there is a [namespace](../concepts/namespace) `files` with
```
aeneasr, line 17: For these links to properly work, you need to append the filename. That helps docusaurus properly resolve the filename. Please do this for all links! You can use search with regexp.

internal/check/engine.go
```diff
@@ -0,0 +1,87 @@
...
+	return &Engine{
+		d: d,
+	}
+}
+
+func (e *Engine) subjectIsAllowed(ctx context.Context, requested *relationtuple.InternalRelationTuple, rels []*relationtuple.InternalRelationTuple) (bool, error) {
+	// This is the same as the graph problem "can requested.Subject be reached from requested.Object through the first outgoing edge requested.Relation"
+	//
+	// recursive breadth-first search
+	// TODO replace by more performant algorithm
```
aeneasr, line 35: Is this tracked as an issue?

internal/check/engine.go
```diff
@@ -0,0 +1,87 @@
...
+
+	var allowed bool
```
aeneasr, line 37:
```suggestion
```

internal/check/engine.go
```diff
@@ -0,0 +1,87 @@
...
+		allowed, err = e.checkOneIndirectionFurther(ctx, requested, &relationtuple.RelationQuery{Object: sub.Object, Relation: sub.Relation, Namespace: sub.Namespace})
+		if err != nil {
+			return false, err
+		}
+		if allowed {
+			return true, nil
+		}
+	}
+
+	return allowed, nil
```
aeneasr, line 61: The default cause is false, not a variable. Remember the macOS TLS vulnerability? ;)

```suggestion
	return false, nil
```

internal/check/engine.go
```diff
@@ -0,0 +1,87 @@
...
+		nextRels []*relationtuple.InternalRelationTuple
+		nextPage string
+	)
+
+	// loop through pages until either allowed, end of pages, or an error occurred
+	for !allowed && nextPage != x.PageTokenEnd && err == nil {
+		nextRels, nextPage, err = e.d.RelationTupleManager().GetRelationTuples(ctx, expandQuery, x.WithToken(nextPage))
+		if errors.Is(err, herodot.ErrNotFound) {
+			allowed, err = false, nil
+			return
```
aeneasr, line 75: This is confusing, why not just

```suggestion
			return false, nil
```

internal/check/engine.go
```diff
@@ -0,0 +1,87 @@
...
+		} else if err != nil {
+			return
```
aeneasr, line 77: Make return values explicit

```suggestion
			return false, err
```

internal/check/engine.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		allowed, err = e.subjectIsAllowed(ctx, requested, nextRels)
```
aeneasr, line 80: Make return values explicit

internal/check/engine.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	return
```
aeneasr, line 82: Make return values explicit, default is deny

```suggestion
	return false, nil
```

internal/check/handler.go
```diff
@@ -0,0 +1,166 @@
...
+type RESTResponse struct {
+	// whether the relation tuple is allowed
+	//
+	// required: true
+	Allowed bool `json:"allowed"`
+}
+
+// swagger:route GET /check read getCheck
+//
+// Check a relation tuple
```
aeneasr, line 68: NYT case for all swagger titles please :)

aeneasr (changes requested):
Can you please re-generate swagger and rebuild the docs so that I can verify the REST API? :)

docs/docs/concepts/subjects.mdx
```diff
@@ -0,0 +1,76 @@
+---
+title: Subjects
+---
+
+import RelationTuplePrism from '@theme/RelationTuplePrism'
+RelationTuplePrism()
+
+In Ory Keto subjects are a recursive polymorphic datatype. They either refer to
+a specific subject (e.g. user) by some application defined identifier, or a set
+of subjects ⭮.
```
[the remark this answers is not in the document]
zepatrik: Haha this was meant as a small little easter egg, it depicts the recursion in this sentence. But I guess it looks like some mistake.

.schema/relation_tuple.schema.json
```diff
@@ -0,0 +1,41 @@
+{
```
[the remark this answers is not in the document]
zepatrik: It is not yet used, I wanted to validate things in the CLI with that but then postponed it. Also, it is useful for editing files like the ones in `contrib/cat-videos-example/relation-tuples`. The IDE will give you code completion and validation.

docs/README.md
```diff
@@ -8,7 +8,7 @@ To change the documentation locally, you need NodeJS installed. Next, install
 the dependencies:
 
 ```
-$ npm
+$ npm install
```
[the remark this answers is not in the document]
zepatrik: This change comes from the template, I just irregularly merged it into this branch.

docs/.widdershins/templates/responses.def
```diff
@@ -12,7 +12,7 @@
 
 |Status|Meaning|Description|Schema|
 |---|---|---|---|
-{{~ data.responses :r}}|{{=r.status}}|{{=r.meaning}}|{{=r.description || 'none'}}|{{=r.schema}}|
+{{~ data.responses :r}}|{{=r.status}}|{{=r.meaning}}|{{ const description = (r.description || 'none').replace('\n', ' '); }}{{=description }}|{{=r.schema}}|
```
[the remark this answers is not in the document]
zepatrik: This change comes from the template, I just irregularly merged it into this branch.

docs/docs/concepts/relation-tuples.mdx
zepatrik: Note to self: this should refer to https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form

internal/check/engine.go
zepatrik: And probably this is caused by the same issue as the macOS vuln: reminiscence of refactoring

docs/docs/concepts/objects.mdx
```diff
@@ -0,0 +1,56 @@
...
+## Basic Example
+
+In the basic case an application uses the same object identifiers as it uses
+internally, e.g. a UUIDv4 like `61e75133-efff-4281-8148-a1806919f568` or SHA-1
+hash like `5c6f593a4e12970d647843f97846fd5ed18179eb`.
+
+Head over to the [basic full feature example](../examples/olymp-file-sharing) to
+see an example with some context.
+
+## Advanced Example: Encoding Application Information within Keto Objects
```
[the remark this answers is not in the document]
zepatrik: Right, I think discouraging that and instead using a UUID that can be mapped to some object by the application is a better idea. Basically like a foreign key to data stored by the application. The actual content can then be updated without any change in Keto.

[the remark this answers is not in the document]
aeneasr: That makes more sense!

docs/docs/concepts/subjects.mdx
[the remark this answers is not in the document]
aeneasr: Ah, I see :D

docs/docs/examples/cat-videos.mdx
zepatrik: I hate docs that have commands defined in a way that I cannot copy-paste them. Syntax highlighting also makes no sense with this format. Are you sure that we want to have them like that?

aeneasr: The problem is that it's not clear what's input and what's output otherwise. Syntax highlighting also works like this if you have `shell-session` as code type. But I agree that it sucks to copy this. However, if it's singular commands, I think it's ok

cmd/version.go
```diff
@@ -1,28 +0,0 @@
-/*
```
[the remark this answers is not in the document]
zepatrik: it is added here: https://github.com/ory/keto/blob/zanzibar/cmd/root.go#L60

cmd/migrate/migrate_test.go
```diff
@@ -0,0 +1,150 @@
...
+				cmd := newMigrateCmd()
+				configx.RegisterFlags(cmd.PersistentFlags())
+				return cmd
+			},
+			Ctx:            ctx,
+			PersistentArgs: persistentArgs,
+		}
+	}
+
+	for _, dsn := range x.GetDSNs(t) {
```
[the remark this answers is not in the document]
zepatrik: Is there a difference between file and in-memory mode?

aeneasr:
Is there a reason why `/relationtuple` is not following REST best practice of plural? Also, is relation tuple one word or two? If it is two, should we use `/relation-tuples`?

aeneasr (commented):
Except for

> Is there a reason why `/relationtuple` is not following REST best practice of plural? Also, is relation tuple one word or two? If it is two, should we use `/relation-tuples`?

and the "todo" pages this is good to go from my side!

aeneasr (approved):
What can I say? Good job!

PR #281, opened 2020-11-03
fix: Handle limits and offsets after filtering 

### LINKED ISSUE #216: Wrong offset behavior when fetching roles with filter by member
**Describe the bug**

The offset works wrong when use filtering by a member in fetch role collection endpoint.

---
The suggestion of cause: Filter applies after roles fetched from DB but offset and limit applied to the DB query without filtering by a member.

**Reproducing the bug**

1. add 2 roles and assign one member to second role.
2. `engines/acp/ory/glob/roles?limit=1&offset=0&member=<member id>` - response is null
3. `engines/acp/ory/glob/roles?limit=1&offset=1&member=<member id>` - response is second role

**Expected behavior**

On step 2 the result must be the second role. But received null

### LINKED ISSUE #280: Limit applied before filtering
**Describe the bug**

Thanks for the great work!

On the endpoint listOryAccessControlPolicies, it looks like the limit of returned policies is applied before filtering by resource or subject. The default limit is 100. If the policies I want are beyond the 100th record, nothing is returned.

**Reproducing the bug**

1. Run keto
2. Insert more policies than the limit returned limit with PUT /engines/acp/ory/exact/policies
3. Fetch the list of Policies, filtering by a subject or resource that is part of the policies insert after the 100th with GET /engines/acp/ory/exact/policies
4. The return is null

If I increase the limit of returned policies, the policy I look for are returned.

_Server configuration_

```yaml
  ory-keto:
    image: oryd/keto:$KETO_VERSION
    container_name: ory-keto
    ports:
      - 4466:4466
    depends_on:
      - ory-hydra
      - ory-keto-migrate
    environment:
      - LOG_LEVEL=$LOG_LEVEL
      - DSN=postgres://$ORY_POSTGRES_USER:$ORY_POSTGRES_PASSWORD@ory-postgres:5432/$ORY_POSTGRES_DB?sslmode=disable
      - AUTHENTICATOR_OAUTH2_INTROSPECTION_CLIENT_ID=$KETO_CLIENT_ID
      - AUTHENTICATOR_OAUTH2_INTROSPECTION_CLIENT_SECRET=$KETO_CLIENT_SECRET
      - AUTHENTICATOR_OAUTH2_INTROSPECTION_TOKEN_URL=$PROTOCOL://$HOST:$PORT/oauth2/token
      - AUTHENTICATOR_OAUTH2_INTROSPECTION_URL=$PROTOCOL://$HOST:$PORT/oauth2/introspect
      - AUTHENTICATOR_OAUTH2_CLIENT_CREDENTIALS_TOKEN_URL=$PROTOCOL://$HOST:$PORT/oauth2/token
    restart: on-failure
```

**Expected behavior**

I expect policies up to the limit, whether they have between inserted beyond the 100th policy or not.

**Environment**

- Version: v0.5.7
- Environment: Docker

xlanor:
## Related issue

Fixes #280 
Fixes #216 

## Proposed changes

I took a look at the source code today because I had some time, and I think that this occurs because in both manager_sql.go and manager.go get the members based on the limit/offset first, before applying the filter in ListByQuery. 

The gist of this PR is that I inspect the query parameters first to see if filters are being applied. If filters are applied, it will retrieve all of the roles/policies, instead of simply the limit/offset specified. Once in memory, I calculate the pagination again in ListByQuery to shorten the array.

## Checklist

      the changes.
      works.

## Further comments

The retrieval from the database (or in-memory datastore) is completely separate from the filtering logic. 
As such, in order to handle this, I needed to retrieve all the possible data, then run the filter on it. After running the filter on it, then I could shorten it by the pagination, in order to ensure that it did not return an empty array.

In addition, I also did a minor refactor on the tests, instead of duplicating all the Roles/Policies in `filter_helper_tests.go`, I simply referred to them from the already declared variable `rolReq` or `polReq` to ensure that updating the role or policy in one place would change across all the rests of the tests. In addition, I also changed the parameters from a `map[string][]string` into a structure, allowing me to specify the offset and limit I wanted to test for each case.


zepatrik (approved):
As discussed, the solution is not ideal, but good enough IMO.

xlanor:
Thank you! @zepatrik . Looking forward to zanzibar!

storage/manager_sql.go
```diff
@@ -117,6 +118,25 @@ func (m *SQLManager) List(ctx context.Context, collection string, value interfac
...
 	}
+	ji := make([]json.RawMessage, len(items))
+	for k, v := range items {
+		ji[k] = json.RawMessage(v)
+	}
+
+	return roundTrip(&ji, value)
+}
+
+func (m *SQLManager) ListAll(ctx context.Context, collection string, value interface{}) error {
```
aeneasr, line 129: Could you add a test for this please? :)

aeneasr (commented):
Awesome, thank you for your contribution! This looks pretty good but I have some ideas how to improve it further :)

aeneasr (approved):

aeneasr:
Awesome 🎉 

PR #313, opened 2020-11-13
chore: make all go packages internal

robinbraemer:
In the meeting @zepatrik and I have come to the decision that making the packages internal to keep our package API slim and later decide what to expose via `pkg` (or other) is reasonable.


aeneasr (approved):
I agree with this!

aeneasr:
@zepatrik I leave merging up to you as it might cause nasty merge conflicts with your ongoing work.

PR #314, opened 2020-11-13
feat: dockerfile - use CMD, disable CGO, use distroless/static image

robinbraemer:
- replace `ENTRYPOINT` with `CMD` to allow overriding the container run command by users
- disable CGO when compiling as its not needed and fixes the common `standard_init_linux.go:211: exec user process caused "no such file or directory"` error developers may face, because of linking to C libs that are not present in slim image
- simply use `GoogleContainerTools/distroless` as base image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,15 @@
 # To compile this image manually run:
 #
-# $ packr; GO111MODULE=on GOOS=linux GOARCH=amd64 go build; docker build -t oryd/keto:latest .; rm keto; packr clean
-FROM alpine:3.9
+# $ packr; GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build; docker build . -t oryd/keto:latest; rm keto; packr clean
 
-RUN apk add -U --no-cache ca-certificates
-
-FROM scratch
-
-COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
-COPY keto /usr/bin/keto
-
+# Use distroless/static as minimal base image that contains:
+# - ca-certificates
+# - A /etc/passwd entry for a root user
+# - A /tmp directory
+# - tzdata
+#
+# Refer to https://github.com/GoogleContainerTools/distroless for more details.
+FROM gcr.io/distroless/static:latest
 USER 1000
-
-ENTRYPOINT ["keto"]
-
-CMD ["serve"]
+COPY keto /usr/bin/keto
+CMD ["keto", "serve"]
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ sdk: deps
 .PHONY: docker
 docker: deps
 		packr
-		GO111MODULE=on GOOS=linux GOARCH=amd64 go build
+		GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build
 		docker build -t oryd/keto:latest .
 		rm keto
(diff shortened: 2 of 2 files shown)


NickUfer:
Nice, didn't know these base images.

> * replace `ENTRYPOINT` with `CMD` to allow overriding the container run command by users

What else than `keto` should a user run in a distroless container?

> "Distroless" images contain only your application and its runtime dependencies. They do not contain package managers, shells or any other programs you would expect to find in a standard Linux distribution.

robinbraemer:
There is not much difference to `ENTRYPOINT` but I used `CMD` in case future developers make changes to the Dockerfile and are aware that they can (should?) use `CMD` to allow overriding the container execute command e.g. when doing `docker run`.

Also if other users make an own image and use `keto` as base, having an `ENTRYPOINT` specified in our Dockerfile may be annoying to people or even block them from basing from us.

zepatrik:
Can we have a build container that has all the build dependencies? Then we can copy the keto binary to a slim `distroless` one. I find it handy to use the shell of a container for debugging. We can still figure out publishing later on.
Also people would not have to install tooling like `pkger` or whatever that would be necessary for builds.
Regarding CGO, we will definitely have to add it later on to a container with sqlite support. This allows for an in-memory database and is nice to start with but also faster for e2e tests.

robinbraemer:
Sure! I knew that would come ^^
I'll move all of the build steps from `make docker` into a build image the next days.

This PR has less priority.

robinbraemer:
> What else than keto should a user run in a distroless container?

@NickUfer To underline my words, I've found an [official example](https://github.com/GoogleContainerTools/distroless/blob/a8b7fa0fc587ec52a0a2ae99e4544907d357e302/examples/go/Dockerfile#L12) where the `CMD` practice is used as well. :+1:

aeneasr (changes requested):
Sorry, this can not be merged as is. Using `entrypoint` is best practice because it allows `docker run oryd/keto --help` for example. You can also override the entrypoint easily using docker CLI flag `--entrypoint`.

Also, we plan on using our trialed and tested build pipeline which supports CGO cross compiling (the error you mentioned simply means that libc is missing or that libc is present but libmusl was expected - which is the case for alpine-based images). We need CGO because we use SQLite extensively for internal in-memory tests and it is fun and easy to use when experimenting with the projects (versus setting up a crdb cluster :) ).

Distroless is also 4x the size of alpine images which is - in my opinion - significant as the binary sizes of our projects are usually below 8mb. Unless there are good reasons why distroless is really that much better than alpine I don't think it makes sense to switch for now.

aeneasr:
As an extension to the above - we always deliver on having non-cgo builds as well which means that you can not use SQLite with these images. The notation is typically `oryd/project:v1.2.3` for non-cgo and `oryd/project:v1.2.3-sqlite` for cgo.

robinbraemer:
Alright, no problem! :+1: 

> ...because we use SQLite extensively for internal in-memory tests and it is fun and easy to use when experimenting with the projects (versus setting up a crdb cluster :) ).

FYI @aeneasr, we can use https://github.com/cockroachdb/cockroach-go/tree/master/testserver for local testing with CRDB as well :)

aeneasr:
Ah nice, that could maybe remove the need for dockertest for CRDB!

However, for internal testing SQLite is probably irreplaceable as it has the best performance because of the way it is used (runs only in memory / no disk writes) and embedded (runs in the same process). :)

Issue #311, opened 2020-11-13
Protobuf APIs & tooling


zepatrik:
**Is your feature request related to a problem? Please describe.**

Currently we use `protoc` wich is quite limited. A better tool might be https://github.com/uber/prototool
It comes with linting.
Also interesting: https://github.com/googleapis/gapic-generator

We might want to use a gateway to avoid writing REST APIs.

robinbraemer:
**Note:** uber/prototool is the "older" tool and bufbuild (also by prototool developers from uber) is the new exciting tool that should be used instead.
- https://github.com/bufbuild/buf

robinbraemer:
Here is the first draft of the APIs: https://github.com/robinbraemer/keto/commit/a2176cc5783ebea78db921958f446cbc36173e60

robinbraemer:
> Ops! Accidentally closed

robinbraemer:
As RBAC and others are planned. I imagine having many different API services. So it might be better to organize `api/` like this:

```
keto/
   acl/
     node/v1/
     admin/v1/
     v1/
   rbac
     node/v1/
     admin/v1/
     v1/
```
...where the `node` package allows for node intercommunication APIs.

(ACL still of cause being the base layer for other models like RBAC.)

robinbraemer:
## "Zookie" renaming suggestions for the Keto project

Since "Zookie" somewhat is a pretty Zanzibar-ish term, I'd like to give it a new name in Keto and that we can define our API appropriately.

**Short recap of what a Zookie is:**
> **2.4 API**
> Zookies in ACL read and check requests specify staleness bounds for snapshot reads, thus providing one of Zanzibar’s core consistency properties. We choose to use an **opaque cookie** instead of the actual timestamp to discourage our clients from choosing arbitrary timestamps and to allow future extensions.

- A Zookie is a "cookie", as in web terminology, replacing the first letter with a `Z` taken from Zanzibar

**This is what I came up with to give "Zookie" its new name in the Keto project:**
- `Kookie`
  - Pretty simple
  - Passes the "Is it a bad word in another language?" - test :D
- `Ketokie`
  - Reasonable
  - Language test: means "Cut away" in Hindi
    - Maybe a lucky coincidence, as zookies have something to do with snapshots and cut away newer information

@zepatrik @aeneasr

zepatrik:
I kinda dislike `kookie` because people might confuse it with actual `cookie`'s and think we have a typo.
[...]

PR #315, opened 2020-11-14
feat: first API draft and generation

robinbraemer:
## Related issue
#311 

Please consider first merging #313.

## Proposed changes

The designs of the APIs are in active discussion with @zepatrik, where we decided that I am the "assignee" who takes more role in maintenance of the APIs because of my former knowledge working with APIs **using** ProtoBuf / gRPC / Client generation.

@zepatrik and I settled down that the two first major parts are **API definitions** and **database schema** that affect development of the ACL system, which makes this PR to get merged ASAP in order to proceed.

After this PR got merged I'll develop on the gapic client generation which users will be using as well as ACL nodes internally for their intercommunications as well as replace the [current protos](https://github.com/ory/keto/blob/zanzibar/relationtuple/relationtuple.proto) so @zepatrik and I can proceed developing using the new protos and the gapic generated clients.

## Access control for Keto client requests itself

As per the discussion with @aeneasr and @zepatrik from the meeting, API endpoints of Keto are only authenticated using mTLS and/**or** authorized through an identity & access proxy (like [oathkeeper](https://github.com/ory/oathkeeper)).

Therefore I decided to split each "access concern" by the following gRCP services, so that administrators can also configure access by using path matching with mesh proxies like in [Istio](https://istio.io/latest/docs/reference/config/networking/virtual-service/)/[Linkerd](https://linkerd.io/2/reference/service-profiles/)/...

Services http paths:
- `/keto.acl.v1.WriteService/<method>`
- `/keto.acl.v1.ReadService/<method>`
- `/keto.acl.v1.CheckService/<method>`
- `/keto.acl.v1.WatchService/<method>`
- `/keto.acl.admin.v1.AdminService/<method>`

Example network policies:
- only allow traffic from very privileged clients to operate on the `AdminService` service.
- only allow traffic from specific clients to access `ReadService` & `CheckService` service.
- only allow traffic from specific clients to access `WatchService` & `AdminService` service.
- etc.

This design does not only allow broad access control by service, but it is also possible to further restrict method access as normal:
  - only allow traffic from specific clients to access `/keto.acl.v1.ReadService/ListRelationTuples` service method.
  - only allow traffic from specific clients to access `/keto.acl.v1.WriteService/DeleteRelationTuples` service method.
  - only allow traffic from specific clients to access `/keto.acl.v1.WatchService/Watch` & `/keto.acl.v1.WriteService/WriteRelationTuples` service methods.
- etc.

Broadly allowing to match the service by the path prefix (e.g. `/keto.acl.v1.WriteService/*`) is more scalable than possible if we would have a mono service (e.g. `AclService`) that would have all read/write/watch/... RPCs in it.

This is because a mono service does not separate concern and would not support administrators to broadly allow clients in a network to a single one access concern. They do also not need to know of every single RPC in advance nor update their network policies in case we add additional RPCs to a service concern.

Opting for a mono service with all RPCs would also open a high security vulnerability because administrators do not know which RPCs we might add in the future.

## Design details / practices
The API takes several other well known API design practices as specified in [AIP](https://google.aip.dev/) (API Improvement Proposals).

  - [Deltas](https://github.com/googleapis/googleapis/blob/be0bdf86cd31aa7c1a7b30a9a2e9f2fd53ee3d91/google/iam/v1/policy.proto#L175) and [States](https://google.aip.dev/154) - for writing / modifying / deleting RelationTuples.
- [Resource freshness validation](https://google.aip.dev/154) - for consistent read-modify-write request flows as specified at 2.4.2 in paper
- [Token/Cursor based Pagination](https://google.aip.dev/158) - for `List...` RPC responses
- [Filtering & Ordering](https://google.aip.dev/132) - for filtering and ordering `List...` RPC responses

## Implementation note
**Note, that** implementing mechanisms such as filtering, ordering or zookie do not need to get implemented right away, but should be the goal to be in in the Keto `v1.0.0` release, where the `v1` APIs gets marked as fixed and stable!


robinbraemer:
I think the first API designs are ready now, so that we can discuss/merge this PR and concentrate on other details, like `Watch API` later! ^^ @zepatrik

robinbraemer:
Maybe it's better to split Write and Delete into simple separate methods before.

robinbraemer:
@zepatrik is the api so okay so far? Do ask me anything that unclear to you. I would also jump on a call.

zepatrik:
I'll review it in ~1h

robinbraemer:
Take your time, no pressure 🍰 😄

api/keto/acl/v1/acl.proto
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Object is an object in a namespace
+// referenced by a RelationTuple or SubjectSet.
+//
+// It represents a "resource" or "digital object".
+message Object {
+  // The namespace of the object.
+  // This effectively is the namespace of the whole RelationTuple.
+  string namespace = 1;
```
zepatrik, line 37: Namespaces will have to have an integer ID that is always constant. This then allows to set an alias (i.e. name) that can be changed. I think it is good to have the alias here, what do you think? So completely fine as you defined it IMO.

api/keto/acl/v1/check_service.proto
```diff
@@ -0,0 +1,71 @@
...
+
+// The request for a CheckService.Check rpc.
+// Checks whether a specific subject is related to an object.
+message CheckRequest {
+  // The related object in this check.
+  Object object = 1;
+  // The relation this check.
+  string relation = 2;
+  // The concrete subject id to check.
+  string subject_id = 3;
```
zepatrik, line 31: In case of a subject set, would this mean that it has to be encoded? Wouldn't it be better if this was of type `Subject`? I am not sure what the use case would be but it is really just a definition thing, so we can allow it without any drawbacks.

api/keto/acl/v1/check_service.proto
```diff
@@ -0,0 +1,71 @@
...
+  //  - your application not yet has a snaptoken on the related data (e.g. user)
+  //
+  // If the specified token is too old, the server falls back and tries to
+  // find the best snapshot token to perform the check using least stale
+  // ACLs that very likely already got replicated.
+  //
+  // *It is recommended to perform checks using slightly stale
+  // data (e.g. token older than 3-10 seconds) for minimum latency
+  // and where the application is allowed to accept slightly off checks.
+  bytes snaptoken = 4;
```
zepatrik, line 50: Maybe it would be better to add them once we have them in the actual code? As they are not strictly required, we can first implement everything without snaptokens/zookies/... and add them in the next step? Do you think it should be in the API definition already anyway?

api/keto/acl/v1/check_service.proto
```diff
@@ -0,0 +1,71 @@
...
+  string subject_id = 3;
+  // Optional. The staleness bound of this check.
+  //
+  // It specifies the clients last known snapshot token
+  // that the server uses to only check against ACLs newer
+  // than specified token snapshot in order to compute the
+  // result of this check.
+  //
+  // Leave this field blank if...
+  //  - your application strictly requires to act on up-to-date data
```
zepatrik, line 40: From the read-api:

> If the request doesn’t contain a zookie, Zanzibar will choose a reasonably recent snapshot, possibly offering a lower latency response than if a zookie were provided.

Do you think the same applies to check or is a check request without a zookie automatically a "content-change" check that returns the new versions zookie? From my understanding it would be possible to have a check without a zookie that then is evaluated at a "reasonably recent snapshot", but that returns no new zookie. Maybe my understanding is wrong here.

api/keto/acl/v1/watch_service.proto
```diff
@@ -0,0 +1,26 @@
...
+option go_package = "github.com/ory/keto/api/keto/acl/v1;acl";
+option csharp_namespace = "Ory.Keto.Acl.V1";
+option java_multiple_files = true;
+option java_outer_classname = "WatchServiceProto";
+option java_package = "sh.ory.keto.acl.v1";
+option php_namespace = "Ory\\Keto\\Acl\\V1";
+
+// The service to watch for changes in the system,
+// such as for Access Control Lists and namespace configs.
+service WatchService {
```
zepatrik, line 17: The watch service will require a lot of background setup first, lets add the API definition together with that functionality.

api/keto/acl/v1/write_service.proto
```diff
@@ -0,0 +1,76 @@
...
+  //
+  // The tuple's etag is used for a consistent read-modify-write request flow.
+  // If the etag is not present, the tuple is not validated against the state
+  // of the serverside tuple and get operated anyway.
+  repeated RelationTupleWriteDelta relation_tuple_deltas = 1;
+}
+
+// Write-delta for a WriteRelationTuplesRequest.
+message RelationTupleWriteDelta {
+  enum Action {
```
zepatrik, line 35: Why would you add this action instead of having different rpcs? Is there a best practice regarding this defined somewhere?

zepatrik (commented):
Nice, I definitely need to learn a bit more about protobuf :sweat_smile:

api/google/api/client.proto
```diff
@@ -0,0 +1,99 @@
+// Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
robinbraemer: Yes, I'll remove and put it back for gapic PR then.

api/keto/README.md
```diff
@@ -0,0 +1,31 @@
+# Notes
+
+> ORY Keto is still a `sandbox` project.
+This makes the included api version `v1` subject
```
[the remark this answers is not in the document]
robinbraemer: Okay, I wasn't sure whether we release next gen keto with a stable v1 directly, but this is not how it goes, since users will be using the service very early I guess.

In protobuf we can use `v1alpha1`, the first alpha of v1, then `v1beta1` and then eventually `v1` for stable definitions.

If we go that path we will always need to write version mappings of the messages versions on server side (map `v1alpha1` -> `v1beta1` -> `v1`), deprecate old versions and eventually remove the support at some point.

Sure, we can do this!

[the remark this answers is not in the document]
zepatrik: Wouldn't it be possible to go with `v0` that can have breaking changes and then move on to `v1` once we release Keto v1.x.x? Otherwise, `v1alpha` and `v1beta` are good :+1:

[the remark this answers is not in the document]
zepatrik: I would remove support for old versions immediately and guarantee compatibility only at a point where we are confident that not much will change.

api/keto/acl/admin/v1/admin_service.proto
```diff
@@ -0,0 +1,20 @@
...
+option go_package = "github.com/ory/keto/api/keto/acl/admin/v1;admin";
+option csharp_namespace = "Ory.Keto.Acl.Admin.V1";
+option java_multiple_files = true;
+option java_outer_classname = "AdminServiceProto";
+option java_package = "sh.ory.keto.acl.admin.v1";
+option php_namespace = "Ory\\Keto\\Acl\\Admin\\V1";
+
+// The admin service for doing administrative tasks in the ACL system.
+service AdminService {
+  option (google.api.default_host) = "keto.exampleapis.com";
```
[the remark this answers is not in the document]
robinbraemer: No, this is the default service host that GAPIC requires to be set for every service. GAPIC generated clients have this host as default for their respective service, but our users would have to pass in their correct API endpoint they deployed and pass it to client. Google uses it to "hard code" their services hostname directly to their clients.

This is not too problematic, I think.
The only error a user would see if he creates a new client and forgets to pass in his correct host would be something like a "hostname keto.exampleapis.com not found".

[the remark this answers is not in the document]
zepatrik: Right, so we add this with GAPIC in a separate PR right? Also, I would prefer to then be something like
```proto
option (google.api.default_host) = "you.have.to.set.hostname.to.your.instance";
```
:wink:

api/keto/acl/admin/v1/admin_service.proto
```diff
@@ -0,0 +1,20 @@
...
+package keto.acl.admin.v1;
+
+import "google/api/client.proto";
+
+option go_package = "github.com/ory/keto/api/keto/acl/admin/v1;admin";
+option csharp_namespace = "Ory.Keto.Acl.Admin.V1";
+option java_multiple_files = true;
+option java_outer_classname = "AdminServiceProto";
+option java_package = "sh.ory.keto.acl.admin.v1";
+option php_namespace = "Ory\\Keto\\Acl\\Admin\\V1";
```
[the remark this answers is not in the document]
robinbraemer: Yes, everything does! :)
As a [sample reference](https://github.com/googleapis/googleapis/blob/3ac5ef0436d8dfeb2ca0091dc7fa8012da1c85af/google/container/v1/cluster_service.proto#L25).

In case, is there a specific `java_package` domain ORY has?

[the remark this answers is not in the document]
zepatrik: Yes, you nailed it already :wink: 
https://github.com/ory/sdk/tree/master/clients/hydra/java#getting-started

api/keto/acl/v1/check_service.proto
robinbraemer: Oh, sure!

I'll put that in, so one can `Check` on a `SubjectSet` (without expantion I guess).

What I just thought is that it might confuse API users if they see they can do a check on a `SubjectSet`.
- I mean one could understand that the semantic is that the server expands the `SubjectSet` to subject ids, therefor making a `Check` request on multiple subject ids. Like WHAT?! :D

robinbraemer: BTW do we also want to add an `skip_expand` bool to the CheckRequest.
- `skip_expand` defaults to `false` so expand checks are done by default (probably what the user expects?)
- setting `skip_expand` skips SubjectSet expanding and any rewrite rules?

api/keto/acl/v1/check_service.proto
robinbraemer: Yes, I do think having them in the API right from the beginning is better.
Users need to learn and adopt the tokens beforehand so they don't need to rewrite their code logics too much.

For now they do not have any special effect since the server always uses the newest tuples.
We could make the server always return a `coming_soon` snaptoken for now. Not an empty one!

api/keto/acl/v1/check_service.proto
zepatrik: Isn't it the same if you expand a `SubjectSet` or not?
And isn't `skip_expand` the same as using read to see whether this exact tuple is stored?

api/keto/acl/v1/check_service.proto
zepatrik: That is a good idea

api/keto/acl/v1/write_service.proto
robinbraemer: The are multiple reasons to this design:
- There would be quite a few and similar looking RPCs to cover each action.
- We can allow clients to do a transaction on multiple tuples in a single RPC call
  - This is very convenient for users in case the user wants all actions to succeed or none!
  - ...while also supporting behaviours like only insert this tuple if it does not already exists, otherwise I want the transaction to fail and give me an error
    - would be pretty easy to implement with SQL

Here are some similar refs:
- [IAM googleapis]( https://github.com/googleapis/googleapis/blob/3ac5ef0436d8dfeb2ca0091dc7fa8012da1c85af/google/iam/v1/policy.proto#L177)
- [GKE googleapis](https://github.com/googleapis/googleapis/blob/75d1324fff65abad6435e6d501e050439b5a616b/google/container/v1/cluster_service.proto#L2018)
- "Actions" can even be whole messages sometimes as in [Spanner googleapis](https://github.com/googleapis/googleapis/blob/b5653ef55579186841dee924592248d821902a12/google/spanner/v1/mutation.proto#L34)

We can always add RPC later if required.

api/keto/acl/v1/watch_service.proto
robinbraemer: I'll remove it.

api/keto/acl/v1/write_service.proto
robinbraemer: Or any other example where 2 or more modifications must succeed.

Imagine this (and I had this scenario learned the hard way):
- Some system inserts 3 tuples in response to an event (e.g. user registration...)
- The system continuously fails at the 3rd insert rpc.
- Now the system would need to rollback somehow and delete the first 2 inserted tuples for clean up.
- ...this is just a mess, cloud be solved client side, but only with much overhead, you get what I mean

- Otherwise, if a transaction is not needed, the user will just make multiple RPCs.

api/keto/acl/v1/check_service.proto
robinbraemer: Right, I'm about to redesign the CheckRequest a bit.

api/keto/acl/v1/acl.proto
robinbraemer: IMO I wouldn't necessarily allow and add aliasing to namespaces. A namespace's FQDN should be just as constant as a primary key.

Why do you want to key a namespace to an integer?

api/keto/README.md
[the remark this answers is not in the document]
robinbraemer: It depends on how long we have the alpha/beta versions public and how many clients are using an older version before should remove them. Generally the goal is to reach v1 as soon as possible.

**This is how versioning is usually seen for ProtoBuf APIs:**
- `v1alpha` - Highly subject to breaking changes, very early stage
- `v1beta` - Pretty reliable, may change. Should have a long evaluation time before going `v1`
- `v1` - Stable, API (RPCs, fields, messages) can only be added without breaking existing functionality

api/keto/acl/admin/v1/admin_service.proto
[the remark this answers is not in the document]
robinbraemer: Right!

api/keto/acl/v1/check_service.proto
robinbraemer: Yes, I mean the expand API (`2.4.5 Expand` in paper), whether if we simply add a `skip_expand` field to the `CheckRequest` that the client can explicitly say whether SubjectSets and Rewrite Rules apply for a check request.

api/keto/acl/v1/check_service.proto
robinbraemer: Done

api/keto/acl/v1/acl.proto
```diff
@@ -0,0 +1,63 @@
...
+  // The object related by this tuple.
+  Object object = 1;
+  // The relation between an Object and a Subject.
+  string relation = 2;
+  // The subject related by this tuple.
+  // A Subject either represents a concrete subject id or
+  // a SubjectSet that expands to more Subjects.
+  Subject subject = 3;
+  // Used to perform a consistent read-modify-write (lock).
+  string etag = 4;
```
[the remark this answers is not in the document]
robinbraemer: An ETAG ([**Resource freshness validation**](https://google.aip.dev/154)) is similar to a hash string of the current version of a RelationTuple including commit_time.
I basically tried to satisfy what is described in paper at `2.4.2 Write`, but maybe in a bit different way?

2.4.2
1) Talks about setting a lock on an object that clients can request in a Read so that in
2) the client does a Write only if that lock had not been changed on server side

I don't quite get it, since this could even result in an "almost infinite" loop between multiple clients competing for the same lock on an object and never manage to run 2 RPCs (read+write) in a row before another client captures the object's lock. Maybe you can help interpret the paper.


...I opted for similar way to support consistent read-modify-write, but **maybe I got lost there** :D
- When the client reads, the server provides the etag, that is the hash of the current RelationTuple snapshot.
- When the client does a write (e.g. delete) on that RelationTuple including the etag (unmodified, as given by server), the server checks that the passed etag equals to the etag computed using the stored RelationTuple to make sure the RelationTuple had not changed in the meantime.

api/keto/acl/v1/write_service.proto
zepatrik: :+1: agree, would really be awesome to have transactional requests here

api/keto/README.md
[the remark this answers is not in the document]
zepatrik: Right, that's what I think as well. So we just start with `v1alpha` and move on when we think it is appropriate.

api/keto/acl/v1/acl.proto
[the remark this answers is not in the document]
zepatrik: So your approach is not locking but rejecting `write` if the object changed between `read` and `write`? That is quite a good idea, as it assures the client that there was no change. On the other hand, tuples have no primary key. So updating one essentially means that we have to delete the old version and add the new version :thinking: I am not sure how to interpret the paper at that point, but I will think about it.

api/keto/acl/v1/acl.proto
zepatrik: I am pretty sure that people will want to rename namespaces... Having an integer ID allows to derive table names from that ID instead of the string name that people would like to change. This way we can have changes to the name but not to the actual table. Anyway, the API should probably use the string name/alias/... anyway, and whether we use integer IDs or not will be part of the namespace config PR.

api/keto/acl/v1/check_service.proto
zepatrik: This would not be a breaking change so let's add it later. Tracked as https://github.com/ory/keto/issues/323

api/keto/acl/v1/acl.proto
robinbraemer: Okey, I see namespaces here quite like a k8s namespace or even database name that does not get changed and is referenced in code and/or configuration. How often would a namespace even change? Giving a namespace a future proof name internal to the service that uses Keto is our users responsibility to get right. Usually namespaces are not even customer-facing so it's pretty internal.

api/keto/acl/v1alpha1/write_service.proto
```diff
@@ -0,0 +1,74 @@
...
+    //
+    // The `WriteRelationTuples` rpc errors if the
+    // specified RelationTuple was not found.
+    UPDATE = 3;
+
+    // Deletion of the RelationTuple.
+    // The `WriteRelationTuples` rpc returns NO error
+    // if the specified RelationTuple was not found.
+    DELETE = 4;
+  }
```
zepatrik, line 60: This only makes sense if we can identify relation tuples somehow, but I don't think it is really possible. An update essentially means
1. delete the old tuple
2. insert the new one

We can add that as a single operation RPC later on, but for now clients can just do these steps themselves. As the parameters will therefore be different for the action types, we should rather separate this into individual RPCs. What do you think?

zepatrik (changes requested):

api/keto/acl/v1alpha1/write_service.proto
robinbraemer: Yes, e.g. the UPDATE action makes little to no sense with the Tuple model.

I'd like to have a single RPC still for transaction support upfront and that we have less Tuple mutation RPCs as possible.

How about having these actions:
- INSERT - no error if already exists
- DELETE - no error if not found

Later we can add actions like:
- INSERT_NEW - errors tx if already exists
- DELETE_FOUND - errors tx if not found

zepatrik: Having a real update that allows deletion and insertion atomically in one call does make sense but is not a hard requirement, so we can add it later. The actions you proposed are perfect and can all be used in one RPC with this action enum.

zepatrik: Tracked as #328 
Should we maybe remove the `etag`s and solve everything consistently? Would be a better approach IMO.

robinbraemer: Yes

robinbraemer:
The snaptoken now is of type string to ease logging/debugging/console output.
The server can use hex string<->bytes mapping.

robinbraemer:
@zepatrik was that it or did I miss something? :)

robinbraemer:
@zepatrik if I did not miss anything, we can squash and merge this baby. :)

zepatrik:
Nice :tada:

PR #331, opened 2020-11-20
feat: incorporate new GRPC API structure

zepatrik:
## Related issue

implements #315


cmd/serve.go
```diff
@@ -68,8 +70,8 @@ on configuration options, open the configuration documentation:
 			defer wg.Done()
 
 			s := grpc.NewServer()
-			relS := relationtuple.NewServer(reg)
-			relationtuple.RegisterRelationTupleServiceServer(s, relS)
+			relS := relationtuple.NewGRPCServer(reg)
+			acl.RegisterReadServiceServer(s, relS)
 			fmt.Println("going to serve GRPC on", lis.Addr().String())
 			if err := s.Serve(lis); err != nil {
```
robinbraemer, line 76: I'll generally have to make some improvements with shutting down the processes, but I'll make another PR after this got merged!

zepatrik: yes, this is not how it should be but it works for first testing

PR #338, opened 2020-11-24
build: pin dependency versions of buf and protoc plugins

zepatrik:
## Related issue

This pins the dependency versions used for protobuf building.

## Proposed changes

Add the dependency build to `make deps` and pin the versions using `go_mod_indirect_pins.go`

## Further comments

For some reason this installs `google.golang.org/protobuf v1.25.1-0.20201020201750-d3470999428b` instead of the latest release `google.golang.org/protobuf v1.25.0`.


zepatrik:
@aeneasr can you have a look if this is done the right way? I figured out that `github.com/bufbuild/buf` has `google.golang.org/protobuf v1.25.1-0.20201020201750-d3470999428b` as a dependency. Therefore we also have to use that although in practice the binaries are independent. Shouldn't we rather use the latest release of `google.golang.org/protobuf v1.25.0`? How would I pin that?

zepatrik:
Sorry @robinbraemer for all the changes, from now on the buf github action will not be used any more. This handles everything, including pinning specific versions, through the Makefile.

PR #383, opened 2021-01-02
feat: expand GPRC service and CLI

zepatrik:
This adds a GPRC service for the expand API.


internal/expand/tree.go
```diff
@@ -85,3 +117,58 @@ func (t *Tree) UnmarshalJSON(v []byte) error {
 
 	return nil
 }
+
+func (t *Tree) ToGRPC() *acl.SubjectTree {
```
robinbraemer, line 121: I'd prefer naming all appearances in the code base of `ToGRPC` and `FromGRPC` to `ToProto`/`FromProto` since it's actually Protocol Buffer messages that are converted, not gRPC messages particularly. Would you rename this in another PR please?

robinbraemer (approved):
I'd prefer naming all appearances in the code base of ToGRPC and FromGRPC to ToProto/FromProto since it's actually Protocol Buffer messages that are converted, not gRPC messages particularly. Would you rename this in another PR please? @zepatrik

robinbraemer (approved):
Otherwise, LGTM!

zepatrik:
Yes, good point. Tracked as #384

PR #440, opened 2021-02-09
fix: empty relationtuple list should not error

### LINKED ISSUE #439: Listing relation tuples where none match cause SQL error
**Describe the bug**

Instead of receiving an empty list, I get: `The requested resource could not be found`.

**Reproducing the bug**

```
	testhelpers.NewKetoServer(t, reg.Config())

	identityID := x.NewUUID()

	projects, err := reg.AccessService().ListAccessibleProjects(ctx, identityID) // could be any non-existing relationtuple
	require.NoError(t, err) // fails with: The requested resource could not be found
	assert.Empty(t, projects)
```

After the first relation has been added, this no longer happens.

zepatrik:
## Related issue
closes #439

### DIFF
--- a/internal/expand/engine.go
+++ b/internal/expand/engine.go
@@ -2,10 +2,6 @@ package expand
 
 import (
 	"context"
-	"errors"
-
-	"github.com/ory/herodot"
-
 	"github.com/ory/keto/internal/x"
 
 	"github.com/ory/keto/internal/relationtuple"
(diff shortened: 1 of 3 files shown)


aeneasr:
Thank you! tests fail though: https://app.circleci.com/pipelines/github/ory/keto/769/workflows/28a3e07f-2df8-4329-8a6d-bbda2822bd98/jobs/6098

Issue #438, opened 2021-02-09
Make errors debuggable and understandable


aeneasr:
**Describe the bug**

Currently, I am failing to list relations or use the check api because of the following error:

```
{"error":{"code":500,"status":"Internal Server Error","message":"rpc error: code = Unknown desc = The requested resource could not be found"}}
```

Neither logs nor anything else give a hint of what's happening and why.

**Reproducing the bug**

_Caller_

```
		res, err := cl.ListRelationTuples(ctx, &acl.ListRelationTuplesRequest{
			Query: &acl.ListRelationTuplesRequest_Query{
				Namespace: relationtuples.ProjectNamespace, // "projects"
				Relation:  relationtuples.ProjectOwnerRelation, // "owner"
				Subject: &acl.Subject{
					Ref: &acl.Subject_Id{Id: identityID.String()},
				},
			},
			ExpandMask: &field_mask.FieldMask{
				Paths: []string{"object"},
			},
			PageSize:  500,
			PageToken: nextPageToken,
		})
```

_Server logs_

```
{"audience":"application","level":"trace","msg":"SELECT EXISTS (SELECT keto_namespace_0000000000_migrations.* FROM keto_namespace_0000000000_migrations AS keto_namespace_0000000000_migrations WHERE version = $1)%!(EXTRA string=20210128140414)","service_name":"ORY Keto","service_version":"master","source":"pop","time":"2021-02-09T12:51:00Z"}
{"audience":"application","level":"trace","msg":"SELECT * FROM keto_0000000000_relation_tuples WHERE relation = $1 AND subject = $2 ORDER BY object, relation, subject, commit_time LIMIT 501 OFFSET 0%!(EXTRA string=owner, string=7920f154-a4f6-4d58-b078-63c68794eda3)","service_name":"ORY Keto","service_version":"master","source":"pop","time":"2021-02-09T12:51:16Z"}
{"audience":"application","level":"trace","msg":"SELECT * FROM keto_0000000000_relation_tuples WHERE relation = $1 AND subject = $2 ORDER BY object, relation, subject, commit_time LIMIT 501 OFFSET 0%!(EXTRA string=owner, string=7920f154-a4f6-4d58-b078-63c68794eda3)","service_name":"ORY Keto","service_version":"master","source":"pop","time":"2021-02-09T12:51:18Z"}
[...]

PR #467, opened 2021-03-05
feat: gRPC status codes and improved error messages

zepatrik:
## Related issue

Contributes to #438 

## Proposed changes

Improves error messages and adds gRPC status codes.


aeneasr:
Is there a best practice around grpc error codes I could read up on or that you have used as basis for this?

zepatrik:
This uses the error codes from https://github.com/ory/herodot/pull/77 where I also linked the documentation.

Issue #466, opened 2021-03-05
Userset Rewrites with Intersections


jon-whit:
The Zanzibar paper mentions that the userset expression language specification supports sub-expressions such as union, intersection, and exclusion.

One use case that I immediately have is to be able to express relation tuples over disjoint sets. In my model I have permission boundaries (analogous to [AWS IAM Permission Boundaries](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_boundaries.html)) and shared/inherited permissions. Permission boundaries specify the maximum level of permissions that a user can have, but it does not specify _what_ permissions the user has. Shared/inherited permissions define the permissions granted to the user either via ABAC policies or explicitly shared rights, but do not fully define the effective permissions the user has.

An access evaluation decision looks like this:

```
boundaries = {"projects.list", "projects.create"}
permissions = {"projects.list"}

 // The effective permissions is the set intersection of the two.
effectivePermissions = boundaries & permissions

effectivePermissions --> {"projects.list"}
```

I'd like to collaborate with you guys on coming up with the namespace configuration expression language that will allow us to express any arbitrary combo of sub-expressions involving unions, intersections, and exclusions (negations).

I've started with the work from [mishudark/zanzibar](https://github.com/mishudark/zanzibar). He's done a pretty good job at laying a foundation down for userset rewrites with unions, but it lacks support for intersection and exclusion. I also don't think that the yaml modeling that he's defined will expand very well once we allow abitrarily nested sub-expressions, so it's probably worth revisiting and going back to the drawing board.

zepatrik:
What about a config similar to this:

![IMG_20210305_174831](https://user-images.githubusercontent.com/5354445/110146682-22cb7100-7ddb-11eb-9ad0-d0165486ba80.jpg)

This is just interesting from a content perspective, not syntax.

jon-whit:
It's less around the configuration syntax and more around the evaluation of such rules.

I've been looking at [Google's Common Expression Language](https://opensource.google/projects/cel) and their go implementation [google/cel-go](https://github.com/google/cel-go). The question is how do we both define the syntax for rewrites and also evaluate arbitrary expressions composing those rewrites for Check and Expand evaluations.

My thinking around using CEL is that our userset rewrite configuration could just translate into a CEL expression, each operand of the expression can be evaluated using the Graph traversal, and then the operands can be fed into the CEL evaluation engine for an outcome.

jon-whit:
From CEL's documentation:

[...]

PR #468, opened 2021-03-05
feat: subject set rewrites with other operator than union

zepatrik:
## Related issue

@jon-whit
#466 

## Proposed changes

This PR contains code changes outlining where the engines would have to be adjusted to allow subject set rewrites with e.g. intersection.

### DIFF
--- a/internal/check/engine.go
+++ b/internal/check/engine.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 
+	"github.com/ory/keto/internal/namespace"
+
 	"github.com/ory/herodot"
 
 	"github.com/ory/keto/internal/relationtuple"
@@ -19,6 +21,7 @@ type (
 	}
 	EngineDependencies interface {
 		relationtuple.ManagerProvider
+		namespace.ManagerProvider
 	}
 )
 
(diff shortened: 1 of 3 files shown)


internal/check/engine.go
```diff
@@ -47,14 +50,24 @@ func (e *Engine) subjectIsAllowed(ctx context.Context, requested *relationtuple.
...
+		nm, err := e.d.NamespaceManager()
+		if err != nil {
+			return false, err
+		}
+		n, err := nm.GetNamespace(ctx, sr.Namespace)
+		if err != nil {
+			return false, err
+		}
+		// use the namespace here to decide which bool operators to use in the loop, and how to adjust the query when moving on (e.g. different relation name)
+		var strategy func(ctx context.Context, requested, current *relationtuple.InternalRelationTuple, checkFurther func(context.Context, *relationtuple.InternalRelationTuple, *relationtuple.RelationQuery) (bool, error)) (continueLoop, allowed bool, err error) = n.GetRelationRewriteStrategy(ctx, sr.Relation)
```
jon-whit, line 62: What does `n.GetRelationRewriteStrategy` do? I don't see that defined in the `namespace/definitions.go`

zepatrik: That is yet to be implemented. In case of a CEL like rewrites it would probably just evaluate the expression.
In case of union the strategy returns true iff one of the subject sets evaluates to true, in case of intersection all would have to, and in case of exclusion it has to evaluate to false.

Until now I returned true at this point iff I found a subject set that returned true. The strategy just implements the logic as required by the config.

jon-whit: Ok, that makes sense at a high level. I'm curious to see that in practice. The challenge I foresee is the nested sub-expressions. Each rewrite expression can involve subexpressions, so these rewrite "strategies" could nest quite deep.

zepatrik: Sure, that is true. But I guess the strategies will be quite simple, at least for now. I am sure that we can optimize it later on. Having the graph search controlled by these strategies will probably be the easiest and most flexible approach.

zepatrik:
This is obsolete with #877

PR #476, opened 2021-03-10
feat: swagger SDK

zepatrik:


zepatrik:
So this is at least now working with swagger, could not get this to work with openapi generator and openapi 3.0 though....
Should we still try to fix that or push that back?

aeneasr:
Push back as we realisticly need kratos 0.6 to land first

PR #486, opened 2021-03-12
docs: suggestions from review

zepatrik:
## Related issue

review in #266

### DIFF
--- a/docs/docs/concepts/objects.mdx
+++ b/docs/docs/concepts/objects.mdx
@@ -8,9 +8,10 @@ RelationTuplePrism()
 Objects are identifiers for some kind of application objects. They can represent
 e.g. a file, network port, physical item, ... . It is up to the application to
 map its objects to an unambiguous identifier. We recommend the usage of UUIDs as
-they provide a high entropy and thus unique identifiers. It is however totally
-possible to use e.g. URLs or opaque tokens of any kind. Ory Keto will consider
-objects equal iff their string representation is equal.
+they provide a high entropy and thus unique identifiers. It is however possible
+to use e.g. URLs or opaque tokens of any kind. Please check the
+[limitations](#advanced-example-using-application-information-within-keto-objects).
+Ory Keto will consider objects equal iff their string representation is equal.
 
 ## Basic Example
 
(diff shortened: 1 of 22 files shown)


docs/docs/concepts/objects.mdx
```diff
@@ -8,49 +8,60 @@ RelationTuplePrism()
...
 
-Because the Keto client can use arbitrary strings as objects, it is possible to
-encode additional information within the object. The only characters not allowed
-within the object string are Ory Keto's delimiters `:#@`. Please refer to the
-documentation of your database for performance considerations, and the maximum
-allowed string length.
+Because the Keto client can use arbitrary strings as objects, it is tempting to
+encode application data within the object. **We strongly discourage this
+practice.** Instead, you should use a UUID to map application data to Keto
+objects. This has a view advantages:
```
aeneasr, line 31:
```suggestion
objects. This has a few advantages:
```

docs/docs/concepts/objects.mdx
```diff
@@ -8,49 +8,60 @@ RelationTuplePrism()
...
 
-For example, this could be used to implement crude checks on value ranges:
+| Advantage                                     | Keto's Limitation                                                         |
```
aeneasr, line 33: I don't really understand this table 😓 What is it supposed to tell us?

docs/docs/concepts/subjects.mdx
```diff
@@ -7,70 +7,80 @@ RelationTuplePrism()
...
+Because the Keto client can use arbitrary strings as subjects, it is tempting to
+encode application data within the subject. **We strongly discourage this
+practice.** Instead, you should use a UUID to map application data to Keto
+subjects. This has a view advantages:
+
+| Advantage                                     | Keto's Limitation                                                         |
+| --------------------------------------------- | ------------------------------------------------------------------------- |
+| control the encoding of data                  | rejects the delimiters `:#@`                                              |
+| control the size of data                      | allows only up to 64 characters                                           |
+| single source of truth <br/> easy data update | stores relation tuples independently of each other <br/> no global update |
```
aeneasr, line 57: Same as in the other document. UUID has advantages? But Keto has limitations? Don't really get it 😅

docs/docs/examples/cat-videos.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: 'Try-Along: Cat Videos'
+title: 'Quickstart: Cat Videos Example'
```
aeneasr, line 2: The document should probably be renamed to docs/docs/quickstart to improve SEO :)

aeneasr (changes requested):
Much better! :) Still have some things that we need to address though :/

docs/docs/concepts/subjects.mdx
zepatrik: I want to tell people why they should use UUIDs instead.

What do you think of

> This is required to ensure
>
> 1. single source of truth and easy data update
> 2. free choice of encoding
> 3. unlimited data size.

aeneasr: Yeah, I think the current table is a bit confusing because it's not clear how the advantage applies to keto's limitation. I think a short list makes probably more sense. If you want, you can add a short description of why this is important in the same bullet point - e.g. `because subjects in Ory Keto can not contain :#@ in their names` or something like that

aeneasr (approved):

PR #487, opened 2021-03-15
test: add command tests

zepatrik:
## Related issue

multiple comments from the review

## Proposed changes

1. add tests for some commands
2. add confirmation question to the migrate commands

### DIFF
--- a/cmd/migrate/down.go
+++ b/cmd/migrate/down.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/ory/x/flagx"
+
 	"github.com/ory/x/cmdx"
 	"github.com/spf13/cobra"
 
@@ -34,6 +36,11 @@ func newDownCmd() *cobra.Command {
 				return cmdx.FailSilently(cmd)
 			}
 			cmdx.PrintTable(cmd, s)
+
+			if !flagx.MustGetBool(cmd, FlagYes) || !cmdx.AskForConfirmation("Do you really want to migrate down? This will delete data.", cmd.InOrStdin(), cmd.OutOrStdout()) {
+				_, _ = fmt.Fprintln(cmd.OutOrStdout(), "Migration aborted.")
+				return nil
+			}
 
 			if err := reg.Migrator().MigrateDown(ctx, int(steps)); err != nil {
 				_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "Could apply down migrations: %+v\n", err)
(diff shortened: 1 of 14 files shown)


cmd/migrate/down.go
```diff
@@ -35,6 +37,11 @@ func newDownCmd() *cobra.Command {
 			}
 			cmdx.PrintTable(cmd, s)
 
+			if !flagx.MustGetBool(cmd, FlagYes) || !cmdx.AskForConfirmation("Do you really want to migrate down? This will delete data.", cmd.InOrStdin(), cmd.OutOrStdout()) {
```
aeneasr, line 40:
```suggestion
			if !flagx.MustGetBool(cmd, FlagYes) || !cmdx.AskForConfirmation("Do you really want to migrate down? This will irrecoverably delete data.", cmd.InOrStdin(), cmd.OutOrStdout()) {
```

cmd/migrate/up.go
```diff
@@ -43,7 +43,7 @@ func newUpCmd() *cobra.Command {
 				return nil
 			}
 
-			if !yes && !cmdx.AskForConfirmation("Do you want to apply above planned migrations?", cmd.InOrStdin(), cmd.OutOrStdout()) {
+			if !yes && !cmdx.AskForConfirmation("Are you sure that you want to apply this migration? Make sure to check the CHANGELOG.md and UPGRADE.md for breaking changes beforehand.", cmd.InOrStdin(), cmd.OutOrStdout()) {
```
aeneasr, line 46: We no longer use UPGRADE.md :)

internal/expand/tree.go
```diff
@@ -130,6 +134,10 @@ func (t *Tree) ToProto() *acl.SubjectTree {
 
 // swagger:ignore
 func TreeFromProto(t *acl.SubjectTree) (*Tree, error) {
+	if t == nil {
+		return nil, nil
```
aeneasr, line 138: Is passing a nil value here not an error?

zepatrik: That depends on whether the expand engine should never return nil. Currently there are two cases:

1. [depth of zero](https://github.com/ory/keto/blob/40563ccb84e966c2feca06c5bc5e247d987afd18/internal/expand/engine.go#L29)
2. [no such relation tuple exists](https://github.com/ory/keto/blob/40563ccb84e966c2feca06c5bc5e247d987afd18/internal/expand/engine.go#L56)

But that can be fixed, although I don't think they are an error.

aeneasr: Ok, as long as the code is able to deal with nil trees downstream I think this is fine. Is that tested?

zepatrik: Not too much (that's why I found it right now), I will add some more tests.

aeneasr (approved):
lgtm from my side

PR #491, opened 2021-03-17
feat: REST patch relation tuples

zepatrik:
The new PATCH handler allows transactional changes similar to the already existing gRPC service.


aeneasr:
Does this implement https://github.com/ory/meta/issues/2 ? If not, PATCH is reserved for patching objects (only change field X) - PUT should be used instead otherwise.

zepatrik:
I am implementing https://tools.ietf.org/html/rfc5789#section-2.1 here. We have an array of objects with the data and action. This enables applying multiple changes in one transaction, just like the gRPC endpoint already does.

aeneasr:
Right, so your are appending or removing items - correct? In that case, `PUT` would be more appropriate: https://en.wikipedia.org/wiki/Patch_verb ( section PUT vs PATCH vs POST):

> The main difference between the PUT and PATCH method is that the PUT method uses the request URI to supply a modified version of the requested resource which replaces the original version of the resource, whereas the PATCH method supplies a set of instructions to modify the resource. If the PATCH document is larger than the size of the new version of the resource sent by the PUT method then the PUT method is preferred.[1]
>
> The POST method can be used for sending partial updates to a resource. The main difference between the POST and PATCH methods is that the POST method can only be used when it is written to support the applications or the applications support its semantics whereas the PATCH method can be used in a generic way and does not require application support. If the outcome of using the PATCH method is not known then the POST method is preferred.[1][6]

zepatrik:
Yes, I am inserting and/or deleting one or more items in one PATCH. The idea is to later add more actions like upsert or update. I do not think that PUT is suitable, as it requires you to transmit the whole resource and not just a part of it. In this case it would mean the whole database. Currently, PUT is used to insert a single item.
I am not sure if POST would work the same way, as my payload literally is a set of instructions - "insert this, delete that, ...". But if you think POST makes more sense we can change that.

Issue #420, opened 2021-02-02
Next Gen Documentation


aeneasr:
- [ ] Quickstart
  - [x] Similar to Ory Hydra or Ory Kratos a quickstart with docker-compose
  - [ ] Explain difference read/write API
- [x] Concepts
  - [x] Namespaces
  - [x] Relation tuples
  - [x] Subjects
  - [x] Subject sets
  - [x] Read & Write APIs
    - [x] Expand API
    - [x] Check API
    - [x] List API
    - [x] Write API
- [ ] Guides
  - [ ] Set up Ory Keto
    - [ ] Explain how to set up namespaces (with config)
    - [ ] Explain most important config settings
    - [ ] Explain how to migrate (if not sqlite)
    - [ ] Explain difference read/write API
    - [ ] Explain how to use the CLI with some samples
  - [ ] Creating a relation tuple using the GRPC API (Write API)
     - [x] Namespace naming convention (singular, upper case, plural, snake case, ...)
     - [x] Object naming convention (if any)
     - [x] Subject naming convention (if any)
     - [x] Relation naming convention (if any)
     - [ ] Explain sharding (naming theory)
     - [ ] Explain Subject Tuples
     - [ ] Create relation tuple
     - [ ] GRPC code example
     - [ ] HTTP code example
  - [x] Check if someone has access to something (Check API)
    - [x] Without indirection
      - [x] GRPC code example
      - [x] HTTP code example
    - [x] With indirection
      - [x] GRPC code example
      - [x] HTTP code example
  - [x] Query relation tuples (to see who has access to e.g. a file) (List API)
      - [x] Explain limitation of indirection here 
      - [x] GRPC code example
      - [x] HTTP code example
  - [x] Expand API
      - [x] Explain what it is for
      - [x] GRPC code example
      - [x] HTTP code example
  - [ ] Managing Namespaces
    - [ ] Configuring
    - [ ] Adding
    - [ ] Removing
  - [ ] Running in Production
    - [ ] Database choice and considerations
    - [ ] Scaling (e.g. just spin up another container)
    - [ ] Applying Migrations
    - [ ] Setting up HTTPS / TLS
  - [ ] Migrating from the legacy Keto
- [x] SDKs
  - [x] GRPC
- [x] Reference
  - [x] REST API
    - [x] Document page_size defaults, limits
    - [x] Document behavior of NextPageToken - e.g. is it set if there are no next results?
  - [x] GRPC API
    - [x] Document page_size defaults, limits
[...]

PR #493, opened 2021-03-18
docs: add API guides

zepatrik:
## Related issue

contributes to #420


zepatrik:
@aeneasr do you want to review this? before merging, the `<CodeTabs version="..." />` has to be replaced. In the current state it works locally with the remote code samples.

Issue #298, opened 2020-11-10
Ensure telemetry is running for GRPC


aeneasr:
**Is your feature request related to a problem? Please describe.**

Figure out how to collect telemetry for GRPC endpoints.

zepatrik:
Only gRPC missing now :tada:

PR #535, opened 2021-04-07
feat: add SQA telemetry

zepatrik:
Contributes to #298 
This adds telemetry for HTTP but not gRPC, as that will require to write a middleware.

### DIFF
--- a/internal/driver/daemon.go
+++ b/internal/driver/daemon.go
@@ -2,6 +2,9 @@ package driver
 
 import (
 	"context"
+	"github.com/ory/keto/internal/check"
+	"github.com/ory/keto/internal/expand"
+	"github.com/ory/keto/internal/relationtuple"
 	"net"
 	"net/http"
 	"strings"
@@ -35,6 +38,10 @@ func (r *RegistryDefault) EnableSqa(cmd *cobra.Command) {
 				healthx.AliveCheckPath,
 				healthx.ReadyCheckPath,
 				healthx.VersionPath,
+
+				relationtuple.RouteBase,
+				check.RouteBase,
+				expand.RouteBase,
 			},
 			BuildVersion: config.Version,
 			BuildHash:    config.Commit,
--- a/internal/driver/daemon.go
+++ b/internal/driver/daemon.go
@@ -6,12 +6,45 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/ory/analytics-go/v4"
+	"github.com/ory/x/healthx"
+	"github.com/ory/x/metricsx"
+	"github.com/spf13/cobra"
+
+	"github.com/ory/keto/internal/driver/config"
+
 	"github.com/ory/graceful"
 	"github.com/pkg/errors"
 	"github.com/soheilhy/cmux"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 )
+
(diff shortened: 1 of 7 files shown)


internal/driver/daemon.go
```diff
@@ -6,13 +6,46 @@ import (
...
+		&metricsx.Options{
+			Service:       "ory-keto",
+			ClusterID:     metricsx.Hash(r.Config().DSN()),
+			IsDevelopment: strings.HasPrefix(r.Config().DSN(), "sqlite"),
+			WriteKey:      "qQlI6q8Q4WvkzTjKQSor4sHYOikHIvvi",
+			WhitelistedPaths: []string{
+				"/",
+				healthx.AliveCheckPath,
+				healthx.ReadyCheckPath,
+				healthx.VersionPath,
```
aeneasr, line 37: Can you add the relation tuple paths here?

internal/driver/daemon.go
```diff
@@ -6,13 +6,46 @@ import (
...
 )
 
+func (r *RegistryDefault) EnableSqa(cmd *cobra.Command) {
+	r.sqaService = metricsx.New(
+		cmd,
+		r.Logger(),
+		r.Config().Source(),
+		&metricsx.Options{
+			Service:       "ory-keto",
+			ClusterID:     metricsx.Hash(r.Config().DSN()),
```
aeneasr, line 30: Are there other values which are kinda unique and stable?

internal/driver/registry_default.go
```diff
@@ -219,6 +222,11 @@ func (r *RegistryDefault) ReadRouter() http.Handler {
 	}
 
 	n.UseHandler(br)
+
+	if r.sqaService != nil {
```
aeneasr, line 226: Is this ever nil? Can we ensure that this doesn't accidentally not run because we forgot to initialize it?

internal/driver/daemon.go
zepatrik: Ah do I have to add all paths?

internal/driver/daemon.go
zepatrik: No not really, maybe the `ListenOn` but that will probably not be very unique.

internal/driver/registry_default.go
zepatrik: It is nil if we do not initialize it.
The problem is, unlike the other projects, we start the server in keto in tests etc. outside of a `cobra.Command` context. That is why I did it this way. There are valid cases where the `sqaService` cannot be initialized.

internal/driver/daemon.go
aeneasr: Yes

internal/driver/daemon.go
aeneasr: Ok then this is enough for now

internal/driver/registry_default.go
aeneasr: Ok, can we add a safe guard to ensure this works when not run using the tests? E.g. via context or something?

zepatrik: What about a `func (r *Registy) ServeWithSQA()` that first adds the service and then serves in one go?

aeneasr: SGTM - but keep in mind that you need the cmd passed down to metricsx.New

aeneasr (approved):

PR #543, opened 2021-04-09
fix: typo

robbietree8:
## Checklist

      the changes.
      works.


zepatrik:
That "iff" is the short form of "if and only if". I guess it is not as well known as I thought. Should we replace it with "if and only if"?

robbietree8:
> That "iff" is the short form of "if and only if". I guess it is not as well known as I thought. Should we replace it with "if and only if"?

omg, when I searched 'iff' in english dictionary, finally got it.

new word learned though..

zepatrik:
Haha np, let's see how many people will also stumble across this and think it is a typo. 2 more and I will change it :wink:

PR #582, opened 2021-05-03
docs: fix example not following best practice and relation tuple syntax highlighting

### LINKED ISSUE #561: Example against best practices
**Describe the bug**

IN https://www.ory.sh/keto/docs/concepts/namespaces it's stated  in the example:

```
// relation describes a relation of the object towards the subject
directories:803a87e9-0da0-486e-bc08-ef559dd8e034#parent@(files:11488ab9-4ede-479f-add4-f1379da4ae43#_)
```

However in  https://www.ory.sh/keto/docs/concepts/graph-of-relations in the example you are doing exactly this bad practice:

```
// Have a look on the subjects concept page if you don't know the empty relation.
dir1#parent@(file1#)
```

should be rather

`dir1#child@(file1#)`

zepatrik:
## Related issue

closes #561

## Further comments

Syntax highlighting broke because docusaurus updated prism and that resulted in https://github.com/FormidableLabs/prism-react-renderer/pull/100 being propagated which in turn removed some of the token classes we used. Added custom styles to the theme to avoid that in the future.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -40,6 +40,8 @@ if (fs.existsSync('./src/css/theme.css')) {
   customCss.push(require.resolve('./src/css/theme.css'))
 }
 
+const githubPrismTheme = require('prism-react-renderer/themes/github')
+
 module.exports = {
   title: config.projectName,
   tagline: config.projectTagLine,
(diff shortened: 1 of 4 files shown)


docs/docusaurus.config.js
```diff
@@ -52,7 +54,54 @@ module.exports = {
   projectName: config.projectSlug, // Usually your repo name.
   themeConfig: {
     prism: {
-      theme: require('prism-react-renderer/themes/github'),
+      theme: {
```
aeneasr, line 57: @zepatrik this will be overwritten by docusaurus-template

zepatrik: I actually thought of adding the styles and language definition to the template, that way we can use it in other docs as well later on.

aeneasr: Feel free to put it in docusaurus-template for now

aeneasr: *as is

PR #583, opened 2021-05-04
feat: make generated gRPC client its own module

zepatrik:
## Related issue

At least two people had issues with using the gRPC client previously, because they had to copy go.mod replacements from our root go.mod
Having it's own go.mod, the gRPC client can now be imported with way less dependencies and without having to apply all those replacements.


aeneasr:
LGTM but nancy is having issues ._.

I think we resolved this for Ory Kratos somehow - maybe need to update the orbs?

zepatrik:
In Kratos we just removed the nancy check completely. I guess it should be fixed now though, let's see :sweat_smile:

Issue #586, opened 2021-05-05
namespace validate not working for keto.yml


arigon:
**Describe the bug**

using `keto namespace validate` only works for isolated namespace files and not for the keto.yml

**Expected behavior**

The validator should recognize only the _namespaces_ property of the keto.yml and validate it.

**Additional context**

```
The configuration contains values or keys which are invalid:
(root)
^-- additionalProperties "version", "dsn", "serve", "profiling", "log", "namespaces" not allowed
```

cramja:
Hi, I've been reading the keto codebase, and I'm curious try this as a first contribution (learning Go in order to read this project)

Intro aside, curious about the expected behavior here. We could detect if a given file is a config file based on the presence of a key called `namespaces` and then parse out the namespaces from there. My guess is that most users will only have a single config file so the command should work like

```
./keto validate config.yaml

OR

./keto validate ns1.yaml ns2.yaml

BUT NOT

./keto validate ns1.yaml config.yaml

AND NOT

./keto validate config.yaml config2.yaml
```

What do you think?

Here's a [hacky branch](https://github.com/cramja/keto/commit/ec2738d93b4ad0a05bc27d9753d9ffe45abaef1a) I used to see how quick it was getting started on this project.

zepatrik:
Awesome, your interpretation looks right. Either, the file is going to be a namespace config file or a root config file containing a pointer to the namespace config file(s) (or the namespace config itself). Duck typing as you proposed seems good :+1: 
Looking forward for your PR :wink:

cramja:
I updated the working branch this morning with code that handles for a config file containing a list of namespace definitions (as opposed to a websocket URL). I will look into adding the websocket validation once I get some more time.

Curious what's the more popular way to configure keto?

zepatrik:
> Curious what's the more popular way to configure keto?

Good question, for me inline because it is less steps. But that is only because the namespace config currently is very limited. With #466 and stuff it will become pretty long at some point.

PR #596, opened 2021-05-15
feat: support config file namespace validation

cramja:
## Related issue

#586 

cc @zepatrik 

## Proposed changes

Adds support for parsing configuration files containing either a list of namespace objects or a reference to a location with namespace objects.

It extends the existing functionality. Here's example usage.

```
### VALID
./keto namespace validate -c config.yaml config2.yaml

./keto namespace validate namespace.yaml ns2.yaml

# namespaces and configs are all validated
./keto namespace validate namespace.yaml ns2.yaml -c config.yaml

### INVALID

# interprets config.yaml as a namespace file
./keto namespace validate config.yaml
```

## Checklist

      the changes.
      works.

## Further comments

I changed the way the namespace watcher works. This is because I wanted to reuse the code in watcher which determines the correct parser type and does an initial scrape of the target location. To make it possible to reuse watcher for validation, I had to leave the original source file contents inside the scraped file.

### DIFF
--- a/cmd/namespace/validate.go
+++ b/cmd/namespace/validate.go
@@ -5,25 +5,43 @@ import (
 	"fmt"
 	"io/ioutil"
 
-	"github.com/ory/keto/internal/driver/config"
-
-	"github.com/ory/x/jsonschemax"
-
 	"github.com/ory/jsonschema/v3"
 	"github.com/ory/x/cmdx"
+	"github.com/ory/x/configx"
+	"github.com/ory/x/jsonschemax"
+	"github.com/ory/x/logrusx"
 	"github.com/segmentio/objconv/yaml"
 	"github.com/spf13/cobra"
 
+	"github.com/ory/keto/internal/driver/config"
 	"github.com/ory/keto/internal/namespace"
 )
 
 func NewValidateCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "validate <namespace.yml> [<namespace2.yml> ...]",
-		Args:  cobra.MinimumNArgs(1),
-		Short: "Validate namespace files",
-		Long:  "Validate one or more namespace yaml files and get human readable errors. Useful for debugging.",
+		Use:   "validate <namespace.yml> [<namespace2.yml> ...] | validate -c <config.yaml>",
+		Short: "Validate namespace definitions",
+		Long: `validate
+Validates namespace definitions. Parses namespace yaml files or configuration
+files passed via the configuration flag. Returns human readable errors. Useful for
+debugging.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			cfFlag := cmd.Flag(configx.FlagConfig)
+			if cfFlag.Changed {
+				cfiles, err := cmd.Flags().GetStringSlice(configx.FlagConfig)
+				if err != nil {
+					_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "Failed to read config command line flag\n%+v\n", err)
+					return cmdx.FailSilently(cmd)
+				}
+				for _, fn := range cfiles {
+					err := validateConfigFile(cmd, fn)
+					if err != nil {
+						return err
+					}
+				}
+			}
+
+			// user passed a list of namespace files
 			for _, fn := range args {
 				_, err := validateNamespaceFile(cmd, fn)
 				if err != nil {
@@ -43,7 +61,7 @@ var configSchema *jsonschema.Schema
 
 const schemaPath = "github.com/ory/keto/.schema/config.schema.json"
 
-func validateNamespaceFile(cmd *cobra.Command, fn string) (*namespace.Namespace, error) {
+func getSchema(cmd *cobra.Command) (*jsonschema.Schema, error) {
 	if configSchema == nil {
 		c := jsonschema.NewCompiler()
 		if err := c.AddResource(schemaPath, bytes.NewBuffer(config.Schema)); err != nil {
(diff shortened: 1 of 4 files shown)


zepatrik (commented):
Awesome, thanks a lot for looking into this 👍 
Did you actually use the namespace watcher changes you made? I could not find how you used it but I am on my phone only so maybe overlooked it 😅

cmd/namespace/validate.go
```diff
@@ -2,28 +2,43 @@ package namespace
...
-		Short: "Validate namespace files",
-		Long:  "Validate one or more namespace yaml files and get human readable errors. Useful for debugging.",
+		Short: "Validate namespace definitions",
+		Long: `validate
+Validates namespace definitions. Parses namespace yaml files or a configuration
+file containing namespace defitions. Returns human readable errors. Useful for
+debugging.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// check user passed a config file
+			if len(args) == 1 {
```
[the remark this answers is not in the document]
cramja: Ah I missed that global flag, but I see it now digging through the code.

I tend to like code where an argument can only be a single type. Since this flag exists, I'd say, let's change the behavior to this style

```sh
### VALID
./keto namespace validate -c config.yaml config2.yaml

./keto namespace validate namespace.yaml ns2.yaml

# namespaces and configs are all validated
./keto namespace validate namespace.yaml ns2.yaml -c config.yaml

### INVALID

# interprets config.yaml as a namespace file
./keto namespace validate config.yaml
```

(the `args` are always treated as namespace files rather than doing the weird type-discovery as this code is now)

What do you think?

internal/driver/config/namespace_watcher.go
```diff
@@ -22,9 +22,20 @@ import (
...
 
 type (
+	parser func([]byte, interface{}) error
+
+	NamespaceFile struct {
+		Name     string
+		Contents []byte
+		Parser   parser
+
+		err       error
```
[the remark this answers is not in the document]
cramja: Yep, on re-read, the err is not necessary to keep around. I'll remove and add a comment that states that the namespace will be nil if it failed to parse

internal/driver/config/namespace_watcher.go
```diff
@@ -167,10 +178,24 @@ func (n *NamespaceWatcher) Namespaces(_ context.Context) ([]*namespace.Namespace
...
+		if nsf.namespace != nil {
+			nscount += 1
+			nspaces = append(nspaces, nsf.namespace)
+		}
 	}
+	return nspaces[0:nscount], nil
+}
 
-	return nspaces, nil
+func (n *NamespaceWatcher) NamespaceFiles() []*NamespaceFile {
```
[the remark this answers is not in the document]
cramja: How about extending a few of the existing tests? (I added those to a new commit - LMK if you'd like altogether new tests)

cmd/namespace/validate.go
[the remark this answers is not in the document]
zepatrik: Sounds perfect :+1:

internal/driver/config/namespace_watcher.go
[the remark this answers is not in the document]
zepatrik: Looks good now :+1:

internal/driver/config/namespace_watcher.go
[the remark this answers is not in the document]
zepatrik: Ah ok I see, maybe it should rather keep the old config if the change update failed. That way nothing breaks while running. In that case it would make sense to have an error and as content the previously successful config. I am just curious where this error is handled in the end.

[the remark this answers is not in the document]
cramja: I see what you mean- that's a good strategy. What I can do is if a namespace fails to parse, I'll update the `Contents` (maybe re-name to `LastReadContents`) but keep the previously parsed namespace.

internal/driver/config/namespace_watcher.go
```diff
@@ -22,12 +22,23 @@ import (
...
+	}
+
 	NamespaceWatcher struct {
 		sync.RWMutex
-		namespaces map[string]*namespace.Namespace
+		namespaces map[string]*NamespaceFile
 		ec         watcherx.EventChannel
 		l          *logrusx.Logger
 		target     string
+		w          watcherx.Watcher
```
cramja, line 41: added for test purposes (force a file reload)

cmd/namespace/validate.go
```diff
@@ -43,44 +64,145 @@ var configSchema *jsonschema.Schema
...
+					_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "Internal error. Failed to marshal yaml %+v\n", err)
+					return cmdx.FailSilently(cmd)
+				}
+				_, err = validateNamespaceBytes(cmd, fmt.Sprintf("index: %d", i), fc, yaml.Unmarshal)
+				if err != nil {
+					return err
+				}
+			}
+			return nil
+		case []map[string]interface{}:
```
cramja, line 188: note: exact same code as above, it just didn't feel right to try to generalize where a simple copy paste seemed to be readable enough. (toml parser returns `[]map[string]interface{}` instead of `[]interface{}`. Interestingly Go does not let you cast `[]map[string]interface{}` as `[]interface{}` which breaks my assumption that `interface{}` stands for anything including `map`)

zepatrik (changes requested):
Sorry for the long time, I made a review but somehow got distracted and did not submit it :facepalm: 
There are just some improvement ideas, but it is nearly good to go :+1:

cmd/namespace/validate.go
cramja: That makes sense. I was thinking that `[]interface{}` expressed "an array of anything" like in Java `List<?>`, but that's not how the compiler sees it

aeneasr:
Just fyi - @zepatrik is on vacation until June 14th so reviews will probably take a bit longer :) I don't really feel that I have enough context to review the PR :/

cmd/namespace/validate.go
```diff
@@ -43,44 +64,145 @@ var configSchema *jsonschema.Schema
...
+	case "yaml", "yml":
+		unmarshal = yaml.Unmarshal
+	case "json":
+		unmarshal = json.Unmarshal
+	case "toml":
+		unmarshal = toml.Unmarshal
+	default:
+		_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "Unhandled config file extension \"%s\"\n", ext)
+		return cmdx.FailSilently(cmd)
+	}
```
zepatrik, line 149: replace this by the function tbd

cmd/namespace/validate.go
```diff
@@ -43,44 +64,145 @@ var configSchema *jsonschema.Schema
...
+					return cmdx.FailSilently(cmd)
+				}
+				_, err = validateNamespaceBytes(cmd, fmt.Sprintf("index: %d", i), fc, yaml.Unmarshal)
+				if err != nil {
+					return err
+				}
+			}
+			return nil
+		default:
+			_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "unknown type for key 'namespace' in config file: %v\n", t)
```
zepatrik, line 202:
```suggestion
			_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "unknown type for key 'namespaces' in config file: %v\n", t)
```

cmd/namespace/validate_test.go
```diff
@@ -0,0 +1,115 @@
...
+const nsyaml = "name: testns0\nid: 0"
+const nsjson = `{"name": "testns1", "id": 1}`
+
+func TestValidateConfigNamespaces(t *testing.T) {
+	cmd := cmdx.CommandExecuter{New: validateCommand}
+
+	t.Run("case=read valid config with embedded namespaces", func(t *testing.T) {
+		dir := t.TempDir()
+		fnyaml := dir + "/keto.yaml"
+		assert.NoError(t, ioutil.WriteFile(fnyaml, []byte(configEmbeddedYaml), fileMode))
```
zepatrik, line 53:
```suggestion
		require.NoError(t, ioutil.WriteFile(fnyaml, []byte(configEmbeddedYaml), fileMode))
```
These are more of a `require`, as it does not make much sense to continue the test if that part fails.

cmd/namespace/validate_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+		stdOut := cmd.ExecNoErr(t, "validate", "-c", fnyaml+","+fnjson+","+fntoml)
+		assert.Equal(t, "Congrats, all files are valid!\n", stdOut)
+	})
+
+	t.Run("case=config passed as varg fails", func(t *testing.T) {
+		fn := t.TempDir() + "/keto.yaml"
+		assert.NoError(t, ioutil.WriteFile(fn, []byte(configEmbeddedYaml), fileMode))
+
+		cmd.ExecExpectedErr(t, "validate", fn)
```
zepatrik, line 69: It would be nice to add some assertions here on the output you get.

cmd/namespace/validate_test.go
```diff
@@ -0,0 +1,115 @@
...
+	})
+
+	t.Run("case=read config with invalid embedded namespace", func(t *testing.T) {
+		fn := t.TempDir() + "/keto.yaml"
+		assert.NoError(t, ioutil.WriteFile(fn, []byte(`{"namespaces":[{"id":"x","name":"x"}]}`), fileMode))
+
+		stdOut := cmd.ExecExpectedErr(t, "validate", "--config", fn)
+		assert.Contains(t, stdOut, "not a valid namespace file")
```
zepatrik, line 77: Also here, it would be nice to at least check if the invalid key is included in the output.
```suggestion
		assert.Contains(t, stdOut, "not a valid namespace file")
		assert.Contains(t, stdOut, "id")
```

internal/driver/config/namespace_watcher.go
```diff
@@ -105,12 +114,19 @@ func eventHandler(ctx context.Context, nw *NamespaceWatcher, done <-chan int, in
 					nw.Lock()
 					defer nw.Unlock()
 
-					n := parseNamespace(nw.l, e.Reader(), e.Source())
+					n := readNamespaceFile(nw.l, e.Reader(), e.Source())
 					if n == nil {
 						return
+					} else if n.namespace == nil {
+						// parse failed
```
zepatrik, line 121:
```suggestion
						// parse failed, rolling back to previous working version
```

zepatrik (changes requested):
I am really sorry about the long review times.
This looks very good, I have some ideas for further improvement.

cmd/namespace/validate.go
```diff
@@ -43,44 +64,145 @@ var configSchema *jsonschema.Schema
 
 const schemaPath = "github.com/ory/keto/.schema/config.schema.json"
 
-func validateNamespaceFile(cmd *cobra.Command, fn string) (*namespace.Namespace, error) {
+func readSchema(cmd *cobra.Command) error {
```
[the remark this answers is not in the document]
cramja: > With this "get or create and cache" function pattern

I like that a lot, thanks!

cmd/namespace/validate_test.go
cramja: Ah, I hadn't encountered the difference before! Java/junit behavior is always `assert`. Will change to require for setup steps.

cramja:
Thanks for the follow up @zepatrik ! Your feedback is very thorough and I appreciate the patience as I learn some Go basics. I will address some more of the advanced suggestions over the next few days and re-request you as a reviewer when complete.

cramja:
@zepatrik re-added you as a reviewer. Thanks again for taking the time for a thorough review. I learned a lot.

Also - I think I will keep the fail-fast strategy for namespace file validation because I think it makes sense to approach the corrections one-by-one.

cmd/namespace/validate.go
```diff
@@ -58,29 +77,121 @@ func validateNamespaceFile(cmd *cobra.Command, fn string) (*namespace.Namespace,
...
+					return cmdx.FailSilently(cmd)
+				}
+				_, err = validateNamespaceBytes(cmd, fmt.Sprintf("index: %d", i), fc, yaml.Unmarshal)
+				if err != nil {
+					return err
+				}
+			}
+			return nil
+		default:
+			_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "unknown type for key 'namespaces' in config file: %v\n", t)
```
zepatrik, line 191: This will print the go type.
Alternatively you can also use `%#v` to print a value including it's type in valid go syntax. 
```suggestion
			_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "unknown type %T for key 'namespaces' in config file: %v\n", t, t)
```

cmd/namespace/validate_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+const nsyaml = "name: testns0\nid: 0"
+const nsjson = `{"name": "testns1", "id": 1}`
+
+func TestValidateConfigNamespaces(t *testing.T) {
+	cmd := cmdx.CommandExecuter{New: validateCommand}
+
+	t.Run("case=read valid config with embedded namespaces", func(t *testing.T) {
+		dir := t.TempDir()
+		fnyaml := dir + "/keto.yaml"
```
zepatrik, line 53: Currently we don't have windows tests, but I think it would be nicer to use
```suggestion
		fnyaml := filepath.Join(dir, "keto.yaml")
```
instead everywhere, so that we could theoretically run this test everywhere.

zepatrik: (applies to all concats below as well, sry :sweat_smile:)

zepatrik (changes requested):
Awesome, this is nearly good to go :heart:

zepatrik:
> Also - I think I will keep the fail-fast strategy for namespace file validation because I think it makes sense to approach the corrections one-by-one.
Good idea btw
Can you then please rebase/merge current master into this?

cramja:
@zepatrik rebased and squashed together!

zepatrik (approved):
Awesome, thanks a lot :heart:

Issue #523, opened 2021-04-02
Allow closing idle connexion after a period


ArthurKnoep:
**Describe the bug**

Currently, the SQL DSN flag `max_conn_lifetime` allow to modify the value passed to `sql.SetConnMaxLifetime` (https://golang.org/pkg/database/sql/#DB.SetConnMaxLifetime), which only close in-use connexion after the period indicated. I think we should also be able to set `sql.SetConnMaxIdleTime` (https://golang.org/pkg/database/sql/#DB.SetConnMaxIdleTime), to be able to close idle connexions after a period.

(The code in question is located in the repo ory/x: https://github.com/ory/x/blob/400477d52ed124d603ce96849ae3bdb549e31b47/sqlcon/connector.go#L186)

I think that we should add a new flag (something like `max_idle_conn_lifetime`) instead of re-using the current `max_conn_lifetime` flag in order to have a more granular configuration to the database connexion pool.

My main motivation for having such a flag is to have Keto close all its connexions to the database after a period of inactivity so that our RDS cluster can scale down.

**Reproducing the bug**

1. Run Keto with postgres
2. In the DSN specify a `max_conn_lifetime` to 30s for example

**Expected behavior**

I expect Keto to close all connexion to the database 30 seconds after the last call to the database 

**Environment**

- Version: v0.5.2
- Postgres version: AWS Aurora Serveless RDS 10.12

**Additional context**

I've only tested this "bug" with Postgres SQL but I assume there will be the same behaviour with MySQL as it use the same underlaying library.

Also this issue follow a discussion on the Ory Slack: https://ory-community.slack.com/archives/C012RBZFMDG/p1617194580015600

ArthurKnoep:
Also if this feature get approved, I'm availaible to implement it

aeneasr:
SGTM, contribs welcomed!

ArthurKnoep:
Ok so for the follow up, I've opened an issue on gobufallo/pop (https://github.com/gobuffalo/pop/issues/632)

Also would it be possible to be assigned on this issue ?

aeneasr:
Of course!

I am also a pop maintainer so could review and merge your PR there :)

ArthurKnoep:
I've opened two PRs, one for `x` as I need to update the internal function that parse the sql flags, and a second one for `keto` that includes the required change to make the new flag work.

I also plan to open a PR for `hydra` in the future

ArthurKnoep:
[...]

PR #605, opened 2021-05-20
feat: max_idle_conn_time

ArthurKnoep:
## Related issue

This PR is related to #523 

## Proposed changes

This PR use the new sql database flag `max_idle_conn_time` to allow configration of https://golang.org/pkg/database/sql/#DB.SetConnMaxIdleTime

## Checklist

      the changes.
      works.

## Further comments

This PR is currently a draft as it needs the merge of ory/x#346


zepatrik:
I guess the only thing missing is updating ory/x to v0.0.242 and it should work?

ArthurKnoep:
> I guess the only thing missing is updating ory/x to v0.0.242 and it should work?

Yep and also updating the doc

I plan to do it tomorrow

ArthurKnoep:
I've updated the dependency and also, it appears that the related docs is not in this repo but in ory/docs.

So this should be ready for a review

zepatrik:
Hm the failing CI is caused by some vulnerable outdated deep-dependency introduced in this PR. How should we handle that @aeneasr ? When using trivy instead of nancy, I find other vulnerabilities...

aeneasr:
If the PR fails sec vuln scanning due to changes it should be fixed in the PR IMO

aeneasr:
For unfound dependencies (e.g. trivy) these should be addressed in the PR that adds the scanner :)

ArthurKnoep:
So, to my understanding the vulnerabilities are caused by this package alone: `golang/k8s.io/kubernetes@1.13.0`, which is declared as `k8s.io/kubernetes` in the go.sum.

However `go mod why` doesn't want to tell why this dependency is needed (from what I understood, `go mod why` only works for direct and indirect dependencies declared in the go.mod)

So what should I do here to fix the vuln ?

zepatrik:
The only way I know of that gives you those dependencies is https://github.com/Helcaraxan/gomod
I checked again and master does not have those alerts. Therefore, they are probably caused by some new dependency added in this PR.

ArthurKnoep:
So, I've finally been able to fix the vuln dependencies. I tried for more than one hour to make `gomod` and `go mod graph | modv` to work with `dot` from graphviz but it appears that there are too many dependencies to draw complete dependency graph in a reasonable amount of time.

So, what I've done is that I've rolled back my changes on go.mod to master, downloaded all the dependencies and cleaned my cache. Then I've re-updated `github.com/gobuffalo/pop/v5` to v5.3.4 and `github.com/ory/x` to v0.0.242 and the vulnerable dependency has now disapeared.
I think it has appeared during the devleopment of my feature and got fixed since I've opened this PR.

We should be good to go now :smile:

aeneasr:
Awesome, thank you so much! :)

aeneasr (approved):

Issue #612, opened 2021-05-31
Importing github.com/ory/keto not compatible with Kratos version 0.6.x


janiskemper:
**Describe the bug**

When importing Keto we need to do the same replacements that are currently in master branch and in version v0.6.0. 

However, the replacement for kratos-client-go is on version v0.5.4-alpha.1.0.20210210170256-960b093d8bf9 which is not compatible with latest Kratos version.

When removing this replacement the error

```
go: github.com/ory/keto@v0.6.0-alpha.3 requires
        github.com/ory/cli@v0.0.49 requires
        github.com/ory/kratos@v0.5.5-alpha.1.0.20210319103511-3726ed4d145a requires
        github.com/ory/cli@v0.0.41 requires
        github.com/ory/kratos@v0.5.5-alpha.1.0.20210210153145-e5a863030890 requires
        github.com/ory/kratos-client-go@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
        
```
occurs. As long as the replacement is there, it is obviously not possible to update kratos-client-go.

**Reproducing the bug**

Just have a Go project importing ```github.com/ory/keto v0.6.0-alpha.3```, use the replacements (otherwise it's not working anyway) and then try to use kratos-client-go on its latest version.

**Expected behavior**

It should be possible to use latest Kratos (0.6.x) and Keto (0.6.x) version together.

aeneasr:
For details see https://github.com/ory/kratos/issues/1363

zepatrik:
Should be closed by #619

janiskemper:
@zepatrik is it possible to make use of your PR? I get an error if I try go get of the latest verified commit:
```
go get github.com/ory/keto@5172ba4e2a0ae76b56d29e924ae8a95131905f4a
go: github.com/ory/keto@v0.6.0-alpha.3 requires
        github.com/ory/cli@v0.0.49 requires
        github.com/ory/kratos@v0.5.5-alpha.1.0.20210319103511-3726ed4d145a requires
        github.com/ory/kratos/corp@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

zepatrik:
Hm I cannot reproduce your error, but I get
```
$ go get github.com/ory/keto@5172ba4e2a0ae76b56d29e924ae8a95131905f4a
go: downloading github.com/ory/keto v0.6.0-alpha.3.0.20210623065416-5172ba4e2a0a
[...]

PR #619, opened 2021-06-15
chore: split "dev dependencies" and real dependencies

zepatrik:
## Related issue

#612 

## Proposed changes

This adds a separate `go.mod` to the `.bin` directory for managing all the binaries required for development/build,
That way we can remove those "dev dependencies" from the package itself.


zepatrik:
Also:
```shell
$ go list -m all | wc -l
539
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
[patrik@patrik-x1 keto]$ go list -m all | wc -l
748
```

aeneasr:
Without `go_mod_indirect_pins.go` `go mod tidy` will remove all dependencies when executed in `.bin`

.bin/go.mod
```diff
@@ -0,0 +1,24 @@
...
+replace github.com/ory/kratos/corp => github.com/ory/kratos/corp v0.0.0-20210118092700-c2358be1e867
+
+replace github.com/oleiade/reflections => github.com/oleiade/reflections v1.0.1
+
+require (
+	github.com/bufbuild/buf v0.31.1
+	github.com/go-swagger/go-swagger v0.26.1
+	github.com/goreleaser/godownloader v0.1.1-0.20200426152203-fd8ad8f7dd78
+	github.com/mattn/goveralls v0.0.7
+	github.com/ory/cli v0.0.49
```
aeneasr, line 18: Install ory/cli using curl bash and the need for any replaces is gone

zepatrik:
> Without go_mod_indirect_pins.go go mod tidy will remove all dependencies when executed in .bin

That's why there is the indirect pins file? https://github.com/ory/keto/pull/619/files#diff-8bd5761c581528cf012968e778e74e9b9c9f219521da0ba7735e1fa5396893bd

.bin/go.mod
zepatrik: Yes, but we will run that more often than necessary. Also, all other go dependencies are still contributing to the dependencies:

> ```shell
> $ go list -m all | wc -l
> 539
> $ git checkout master
> Switched to branch 'master'
> Your branch is up to date with 'origin/master'.
> $ go list -m all | wc -l
> 748
> ```

This PR reduces the dependency count by ~28%.

aeneasr:
Exactly!

.bin/go.mod
aeneasr: True, that's definitely nice!

aeneasr (approved):

Issue #628, opened 2021-06-21
Migration path for v0.7


zepatrik:
**Is your feature request related to a problem? Please describe.**

#613 will internally break compatibility with v0.6. We have to provide a migration path that needs to go through go code for parsing subject sets.

**Describe the solution you'd like**

A CLI command that migrates the data.

Issue #618, opened 2021-06-14
SQL migration tests


zepatrik:
Migration tests are missing. Add similar to https://github.com/ory/kratos/tree/master/persistence/sql/migratest

Issue #599, opened 2021-05-17
Bulk deletion of relation tuples


zepatrik:
**Is your feature request related to a problem? Please describe.**

To allow full RBAC #598, we have to delete multiple (might be thousands) relation tuples at once. Another use case is where a user has to be fully deleted (think GDPR and friends) or general cleanup tasks. Keeping relation tuples for deleted objects/users might not be desirable or even a bug if another object is created with the same identifier at some later point.

**Describe the solution you'd like**

Allow specifying partial relation tuples similar to the list API for deletion.

**Describe alternatives you've considered**

Keep current behavior and require clients to iterate over all relation tuples and deleting them.

vancanhuit:
Hey, I can take this issue.

zepatrik:
Good, the idea is basically to allow supplying partial tuples just like the list API. Maybe you can even reuse it's SDK definitions.

PR #638, opened 2021-06-29
refactor: persistence table structure

### LINKED ISSUE #448: Use fizz migrations to generate SQL migrations
**Describe the bug**

The SQL migrations are currently handwritten and ignore the fizz templating pipeline we've built. The pipeline should be used instead.

### LINKED ISSUE #613: Rework persistence to not have a table per namespace

Currently we have tables per namespace. This decision was made based on the idea to have storage specific settings (like additional indexes) per namespace (outlined in #303). The problem with that is that it is not scaling well in terms of maintenance and performance in some cases.

A much more scalable solution will be to use a single table that has a namespace column. We might still be able to move back to separate tables by implementing another persister, but that is a topic for when the APIs are considered stable.

**Additional context**

from #303 

> Google Zanzibar §2.3
> 
> The paper specifies the following storage parameters:
> 
> > Storage pa-rameters include sharding settings and an encoding for objectIDs that helps Zanzibar optimize storage of integer, string,and other object ID formats.
> 
> From my first experience with the system, it will be helpful to allow configure database indexes as well, as the read-API can be used in different ways by a client.

https://github.com/cockroachdb/cockroach/issues/63206

zepatrik:
## Related issue

closes #613 
closes #448 

## Proposed changes

As outlined in #613 we now have one table for all relation tuples:

```sql
CREATE TABLE keto_relation_tuples
(
    shard_id                 UUID        NOT NULL,
    network_id               UUID        NOT NULL,
    namespace_id             INTEGER     NOT NULL,
    object                   VARCHAR(64) NOT NULL,
    relation                 VARCHAR(64) NOT NULL,
    subject_id               VARCHAR(64) NULL,
    subject_set_namespace_id INTEGER NULL,
    subject_set_object       VARCHAR(64) NULL,
    subject_set_relation     VARCHAR(64) NULL,
    commit_time              TIMESTAMP   NOT NULL,

    PRIMARY KEY (shard_id, network_id),

    -- enforce to have exactly one of subject_id or subject_set
    CONSTRAINT chk_keto_rt_subject_type CHECK
        ((subject_id = null AND
          NOT subject_set_namespace_id = null AND NOT subject_set_object = null AND NOT subject_set_relation = null)
            OR
         (NOT subject_id = null AND
          subject_set_namespace_id = null AND subject_set_object = null AND subject_set_relation = null))
);
```

## Checklist

- [x] I am following the
      the changes.
      works.

## Further comments

Outstanding are still
#628 
#618 
but those will be separate PRs.


Makefile
```diff
@@ -16,12 +15,15 @@ GO_DEPENDENCIES = github.com/go-swagger/go-swagger/cmd/swagger \
...
   # go install is responsible for not re-building when the code hasn't changed
-  .bin/$(notdir $1): go.mod go.sum Makefile
+  .bin/$(notdir $1): .bin/go.mod .bin/go.sum Makefile
 		cd .bin; GOBIN=$(PWD)/.bin/ go install $1
 endef
 $(foreach dep, $(GO_DEPENDENCIES), $(eval $(call make-go-dependency, $(dep))))
 $(call make-lint-dependency)
 
+.bin/ory: .bin/go.mod .bin/go.sum Makefile
+		cd .bin; GOBIN=$(PWD)/.bin/ go install -tags sqlite github.com/ory/cli; mv cli ory
```
zepatrik, line 25: separate task because of `-tags sqlite`

cmd/migrate/migrate_test.go
```diff
@@ -126,24 +125,24 @@ func TestMigrate(t *testing.T) {
...
+				//	parts := regexp.MustCompile("(?s)Current status:(.*)Successfully applied all migrations(.*)Going to migrate namespaces(.*)Successfully applied all migrations(.*)Current status(.*)Successfully applied all migrations(.*)").FindStringSubmatch(out)
+				//	require.Len(t, parts, 7)
+				//
+				//	assertNoneApplied(t, parts[1])
+				//	assertAllApplied(t, parts[2])
+				//	assertNoneApplied(t, parts[3])
+				//	assertAllApplied(t, parts[4])
+				//	assertNoneApplied(t, parts[5])
+				//	assertAllApplied(t, parts[6])
+				//})
```
zepatrik, line 145: this can probably be removed

cmd/namespace/migrate_down.go
```diff
@@ -1,56 +1,21 @@
...
 	"github.com/spf13/cobra"
-
-	"github.com/ory/keto/internal/driver"
 )
 
 func NewMigrateDownCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "down <namespace-name> <steps>",
-		Short: "Migrate a namespace down",
+		Deprecated: "This step is not necessary anymore, see TODO",
```
zepatrik, line 11: Add link to this PR or changelog here

cmd/namespace/migrate_status.go
```diff
@@ -1,45 +1,19 @@
...
-				return cmdx.FailSilently(cmd)
-			}
-
-			mb, err := reg.NamespaceMigrator().NamespaceMigrationBox(cmd.Context(), n)
-			if err != nil {
-				return err
-			}
-
-			return migrate.BoxStatus(cmd, mb, "[namespace="+n.Name+"] ")
+		Deprecated: "This step is not necessary anymore, see TODO",
```
zepatrik, line 10: same

cmd/namespace/migrate_up.go
```diff
@@ -1,48 +1,20 @@
...
-				return cmdx.FailSilently(cmd)
-			}
-
-			mb, err := reg.NamespaceMigrator().NamespaceMigrationBox(ctx, n)
-			if err != nil {
-				return err
-			}
-
-			return migrate.BoxUp(cmd, mb, "[namespace="+n.Name+"] ")
+		Deprecated: "This step is not necessary anymore, see TODO",
```
zepatrik, line 11: same

internal/persistence/sql/persister.go
```diff
@@ -56,7 +59,7 @@ var (
 )
 
 func NewPersister(dsn string, l *logrusx.Logger, namespaces namespace.Manager, tracer *tracing.Tracer) (*Persister, error) {
-	pop.SetLogger(l.PopLogger)
+	pop.Debug = true
```
zepatrik, line 62: revert this

cmd/namespace/migrate_down.go
```diff
@@ -1,56 +1,21 @@
...
 	"github.com/spf13/cobra"
-
-	"github.com/ory/keto/internal/driver"
 )
 
 func NewMigrateDownCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "down <namespace-name> <steps>",
-		Short: "Migrate a namespace down",
+		Deprecated: "This step is not necessary anymore, see https://github.com/ory/keto/pull/638",
```
aeneasr, line 11: I think it's fine to delete this command

cmd/namespace/migrate_status.go
```diff
@@ -1,45 +1,19 @@
...
-				return cmdx.FailSilently(cmd)
-			}
-
-			mb, err := reg.NamespaceMigrator().NamespaceMigrationBox(cmd.Context(), n)
-			if err != nil {
-				return err
-			}
-
-			return migrate.BoxStatus(cmd, mb, "[namespace="+n.Name+"] ")
+		Deprecated: "This step is not necessary anymore, see https://github.com/ory/keto/pull/638",
```
aeneasr, line 10: I think it's fine to delete this command

cmd/namespace/migrate_up.go
```diff
@@ -1,48 +1,20 @@
...
-				return cmdx.FailSilently(cmd)
-			}
-
-			mb, err := reg.NamespaceMigrator().NamespaceMigrationBox(ctx, n)
-			if err != nil {
-				return err
-			}
-
-			return migrate.BoxUp(cmd, mb, "[namespace="+n.Name+"] ")
+		Deprecated: "This step is not necessary anymore, see https://github.com/ory/keto/pull/638",
```
aeneasr, line 11: I think it's fine to delete this command

internal/driver/config/provider.go
```diff
@@ -131,66 +134,68 @@ func (k *Provider) CORS() (cors.Options, bool) {
...
+	if nm := func() namespace.Manager {
+		k.nmLock.RLock()
+		defer k.nmLock.RUnlock()
+		return k.nm
+	}(); nm != nil {
+		return nm, nil
+	}
+
+	fmt.Println("creating new nm")
+
```
aeneasr, line 167:
```suggestion
```

internal/driver/config/provider.go
```diff
@@ -131,66 +134,68 @@ func (k *Provider) CORS() (cors.Options, bool) {
...
+	k.nmLock.Lock()
+	defer k.nmLock.Unlock()
+
+	var ctx context.Context
+	ctx, k.cancelNamespaceManager = context.WithCancel(k.ctx)
```
aeneasr, line 172: Do you have a test how this behaves at scale or with constant reloading? We should avoid any context or memory leaks.

internal/persistence/sql/network_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+				t.Run("case=gets same ID with new persister", func(t *testing.T) {
+					ctx, reg := setup(t, dsn)
+
+					nid2, err := reg.Persister().NetworkID(ctx)
+					require.NoError(t, err)
+					assert.Equal(t, nid, nid2)
+				})
+
+				t.Run("case=db enforces single network ID", func(t *testing.T) {
```
aeneasr, line 74: I would remove this

internal/persistence/sql/network.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (*NetworkID) TableName() string {
+	return "keto_networks"
+}
+
+func (p *Persister) NetworkID(ctx context.Context) (uuid.UUID, error) {
+	if p.networkIDCached == uuid.Nil {
+		var nID NetworkID
+
+		if err := p.Connection(ctx).First(&nID); errors.Is(err, sql.ErrNoRows) {
```
aeneasr, line 30: Please use `networkx.Manager.Determine()`

internal/relationtuple/manager_requirements.go
```diff
@@ -44,10 +44,12 @@ func ManagerTest(t *testing.T, m Manager, addNamespace func(context.Context, *te
 			for _, tup := range tuples {
```
aeneasr, line 44: Please add tests that ensure network isolation

aeneasr (commented):
So how would a migration from the old model to the new model look like? Given that we need that in prod and stage

.bin/go.mod
```diff
@@ -4,18 +4,29 @@ go 1.16
 
 replace github.com/goreleaser/nfpm => github.com/goreleaser/nfpm v1.10.2
 
-replace github.com/ory/kratos-client-go => github.com/ory/kratos-client-go v0.5.4-alpha.1.0.20210210170256-960b093d8bf9
+replace github.com/ory/kratos-client-go => github.com/ory/kratos-client-go v0.6.3-alpha.1
```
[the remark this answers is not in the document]
zepatrik: This is in`.bin` so that we don't have the dependency in the main module. But also I don't want to always wait for the download, when go can rebuild stuff just on changes... But if you think that makes more sense, please release https://github.com/ory/cli/pull/90 or can I just make a release?

cmd/namespace/migrate_down.go
zepatrik: If we instead we add the deprecated text, someone using it will now what is going on. We can delete it at a later point. A deprecated command does not show up in the help or docs.

internal/driver/config/provider.go
zepatrik: this code was just moved from https://github.com/ory/keto/pull/638/files#diff-27444dce264cbc384993274b30b80fcf763ad3255d0a7df124d20d2e047dad21L159-L160

cmd/namespace/migrate_down.go
aeneasr: True

internal/driver/config/provider.go
aeneasr: Question still remains :)

.bin/go.mod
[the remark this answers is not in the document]
aeneasr: The touch command will remove any uneccessary downloads! I'll create a release immediately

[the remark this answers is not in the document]
zepatrik: once you change the makefile it re-downloads the binary?

internal/driver/config/provider.go
zepatrik: No there is no specific test for this context. Tracked as #642

.bin/go.mod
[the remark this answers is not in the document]
aeneasr: Release is in the making!

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.up.sql
```diff
@@ -0,0 +1,25 @@
+-- we can not fizz this migration because there is no check constraint support in fizz
+-- and it gets cumbersome with sqlite; having a working SQL version is actually way easier
+CREATE TABLE keto_relation_tuples
+(
+    shard_id                 UUID        NOT NULL,
+    network_id               UUID        NOT NULL,
```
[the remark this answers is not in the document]
zepatrik: Just realized that I did not use `nid` because we also have a namespace ID and that is not descriptive enough to distinguish them... but maybe better to stay consistent still.

cmd/server/serve.go
```diff
@@ -37,8 +43,11 @@ on configuration options, open the configuration documentation:
 
 >> https://www.ory.sh/keto/docs/reference/configuration <<`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			reg, err := driver.NewDefaultRegistry(cmd.Context(), cmd.Flags())
-			if err != nil {
+			reg, err := driver.NewDefaultRegistry(cmd.Context(), cmd.Flags(), false)
+			if errors.Is(err, persistence.ErrNetworkMigrationsMissing) {
+				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Migrations were not applied yet, please apply them first.")
```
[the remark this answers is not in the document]
zepatrik: probably yes

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.mysql.up.sql
```diff
@@ -0,0 +1,25 @@
+-- we can not fizz this migration because there is no check constraint support in fizz
+-- and it gets cumbersome with sqlite; having a working SQL version is actually way easier
+CREATE TABLE keto_relation_tuples
+(
+    shard_id                 char(36)    NOT NULL,
+    nid                      char(36)    NOT NULL,
+    namespace_id             INTEGER     NOT NULL,
```
aeneasr, line 7: Wouldn't this compose the primary key also? And if so, would it make sense to assign it a UUID? An int might become difficult in multi-tenant scenarios where you have lots of namespaces in lots of nids

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.mysql.up.sql
```diff
@@ -0,0 +1,25 @@
...
+    PRIMARY KEY (shard_id, nid),
+
+    -- enforce to have exactly one of subject_id or subject_set
+    CONSTRAINT chk_keto_rt_subject_type CHECK
+        ((subject_id IS NULL AND
+          subject_set_namespace_id IS NOT NULL AND subject_set_object IS NOT NULL AND subject_set_relation IS NOT NULL)
+            OR
+         (subject_id IS NOT NULL AND
+          subject_set_namespace_id IS NULL AND subject_set_object IS NULL AND subject_set_relation IS NULL))
+);
```
aeneasr, line 25: Aren't we lacking some indices here? We probably have several quries with e.g. `subject_id` involved. These should be index IMO.

internal/persistence/sql/relationtuples.go
```diff
@@ -138,16 +250,34 @@ func (p *Persister) GetRelationTuples(ctx context.Context, query *relationtuple.
...
+			return nil, "", err
+		}
+		wheres = append(wheres, whereStmts{stmt: "namespace_id = ?", arg: n.ID})
 	}
 
-	sqlQuery := p.connection(context.WithValue(ctx, namespaceContextKey, n)).
-		Order("object, relation, subject, commit_time").
+	sqlQuery := p.Connection(ctx).
+		Where("nid = ?", p.NetworkID(ctx)).
+		Order("nid, namespace_id, object, relation, subject_id, subject_set_namespace_id, subject_set_object, subject_set_relation").
```
aeneasr, line 280: These are quite a lot of order operations. Does that really make sense? These would all need to be indexed.

Also, ordering by `nid` doesn't seem logical to me as `nid` is always the same value for every row

internal/persistence/sql/relationtuples.go
```diff
@@ -18,105 +20,210 @@ import (
...
+	); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (p *Persister) deleteRelationTupleSubjectID(ctx context.Context, r *relationtuple.InternalRelationTuple, s *relationtuple.SubjectID, n *namespace.Namespace, network uuid.UUID) error {
+	if err := p.Connection(ctx).RawQuery(
+		"DELETE FROM keto_relation_tuples WHERE nid = ? AND namespace_id = ? AND object = ? AND relation = ? AND subject_id = ?",
+		network,
```
aeneasr, line 165: Shouldn't this be `p.NetworkID(ctx)` to avoid fat-finger issues?

internal/persistence/sql/relationtuples.go
```diff
@@ -18,105 +20,210 @@ import (
...
-			ID:         shardID,
-			Object:     rel.Object,
-			Relation:   rel.Relation,
-			Subject:    rel.Subject.String(),
-			CommitTime: time.Now(),
-		}),
-	)
+	if err := p.Connection(ctx).RawQuery(
+		"DELETE FROM keto_relation_tuples WHERE nid = ? AND namespace_id = ? AND object = ? AND relation = ? AND subject_set_namespace_id = ? AND subject_set_object = ? AND subject_set_relation = ?",
+		network,
```
aeneasr, line 190: Shouldn't this be `p.NetworkID(ctx)` to avoid fat-finger issues?

internal/persistence/sql/relationtuples.go
```diff
@@ -18,105 +20,210 @@ import (
...
-		p.connection(context.WithValue(ctx, namespaceContextKey, n)).Create(&relationTuple{
-			ID:         shardID,
-			Object:     rel.Object,
-			Relation:   rel.Relation,
-			Subject:    rel.Subject.String(),
-			CommitTime: time.Now(),
-		}),
-	)
+	if err := p.Connection(ctx).RawQuery(
+		"DELETE FROM keto_relation_tuples WHERE nid = ? AND namespace_id = ? AND object = ? AND relation = ? AND subject_set_namespace_id = ? AND subject_set_object = ? AND subject_set_relation = ?",
```
aeneasr, line 189: I assume it is not possible to do this by the shard ID?

internal/persistence/sql/relationtuples.go
```diff
@@ -18,105 +20,210 @@ import (
...
 
-			if err := c.RawQuery(
-				fmt.Sprintf("DELETE FROM %s WHERE object = ? AND relation = ? AND subject = ?", tableFromNamespace(n)),
-				r.Object,
-				r.Relation,
-				r.Subject.String(),
-			).Exec(); err != nil {
-				return sqlcon.HandleError(err)
+			switch s := r.Subject.(type) {
+			case *relationtuple.SubjectID:
```
aeneasr, line 217: Would it maybe make sense to find the element, and then delete it? So that we have only one place which constructs the (quite) complex search query?

internal/relationtuple/manager_requirements.go
```diff
@@ -309,6 +328,36 @@ func ManagerTest(t *testing.T, m Manager, addNamespace func(context.Context, *te
 			require.NoError(t, err)
 			assert.Equal(t, []*InternalRelationTuple{rs[1], rs[3]}, res)
 		})
+
+		t.Run("case=tuple and subject namespace differ", func(t *testing.T) {
```
aeneasr, line 332: So it is possible to have a relation in a different namespace and link it to a subject in another namespace? Would that cause issues when we delete something something in namespace `2` on which something in namespace `3` depends on?

aeneasr (changes requested):
I found a few things:

* [ ] Files `20201110175414_relationtuple` do not follow the new strategy. I think appending trailing `0` should be enough.
* [ ] Missing index for `DELETE FROM keto_relation_tuples WHERE nid = ? AND namespace_id = ? AND object = ? AND relation = ? AND subject_id = ?`
* [ ] Missing index for `DELETE FROM keto_relation_tuples WHERE nid = ? AND namespace_id = ? AND object = ? AND relation = ? AND subject_set_namespace_id = ? AND subject_set_object = ? AND subject_set_relation = ?`
* [ ] Missing index for `GetRelationTuples()`

And a few more questions:

- What is the difference between `RelationQuery` and `InternalRelationTuple`?

Also we do not have network isolation tests yet. Do you want to include that in this PR or in another one?

internal/persistence/sql/relationtuples.go
zepatrik: The shard ID is completely internal and never exposed, so we cannot get it as part of a request.

internal/relationtuple/manager_requirements.go
zepatrik: Yes, this is by design. With the new schema we can actually query for those cases as well when deleting, but I am not sure if it makes much sense to actually delete them; ex:
```
groups:admins#member@aeneas
files:/foo#delete@(groups:admins#member)
```
If we delete the first tuple, I don't think you want to delete the second one, even though currently there is no admin.

The de-serialized table structure now actually allows us to query e.g. for tuples referring to nothing, but you might actually want to keep those data anyway.

internal/persistence/sql/relationtuples.go
zepatrik: To get reliable pagination we have to sort things. The `nid` can probably be removed, but I think the idea was to use index prefixes to minimize the amount of (yet missing) indexes. Will add them though and check whether the `nid` has to be included or not.

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.mysql.up.sql
zepatrik: Yes, but why add the `namespace_id` to the PK?
The namespace ID is an int because it is managed by the user (maybe add docs for that :sweat_smile:). It determines the mapping between the namespace name (used in the API) and the actual database entry. This allows easy renaming of the namespaces as renaming does not affect the database at all. It was like that already before, is there a reason to make this a breaking change? Would you then create the UUIDs internally? How do you then handle changes to the namespace?

internal/persistence/sql/relationtuples.go
zepatrik: What do you mean by find it? Use `GetRelationTuples` and then use the `shard_id` for deletion instead? How do you then handle duplicate tuples? Get all of the matching tuples, maybe even paginated, and delete with `shard_id = ? OR shard_id = ? ...`?

internal/relationtuple/manager_isolation.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Run(desc+"-0", func(t *testing.T) {
+			run(t, m0, m1)
+		})
+		t.Run(desc+"-1", func(t *testing.T) {
+			run(t, m1, m0)
+		})
+	}
+}
+
+func IsolationTest(t *testing.T, m0, m1 Manager, addNamespace func(context.Context, *testing.T, string)) {
```
zepatrik, line 22: Regarding isolation, what cases are missing in this test in your opinion @aeneasr? Or how should the isolation test work instead of this?

zepatrik:
>  Files 20201110175414_relationtuple do not follow the new strategy. I think appending trailing 0 should be enough.

I can't change the migration ID for already applied migrations right? That would break existing installations as far as I can tell.

---

> Missing index for GetRelationTuples()

Absolutely correct, but I would have to index everything in various orders to match all cases. Depending on the application needs you have different queries. That was the whole point of why we had tables per namespace until now, so that we could add indexes matching the application needs, and not having to index everything. IMO we should right now have some slow queries whenever the indexes are not there, but add the ones required for `check` evaluation. Later we can add configuration to the namespaces that would enable partial indexes, so that we can say in the `groups` namespace, we want to query by object and relation (think: listing all members), while in the `files` namespace we want to query by subject and relation (think: listing all files a user is the owner of).

TL;DR

I will add indexes required for the `check` and `expand` APIs, but not for user-defined queries.

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.mysql.up.sql
aeneasr: Yeah, the problem in Ory Cloud will be that we have e.g. 100.000 namespaces, and potentially several customers creating namespaces concurrently (specifically across regions) which would mean that we have to do a table lock in order to fetch the highest ID (basically the auto_increment issue) on potentially billions of elements (as we have all relations in here). So it would be better to have something unique here, IMO.

I don't know how that could be solved in the user config. Potentially with some additional table / mapping:

```
struct NamespaceMapping {
  ConfigKey int
  ID uuid.UUID
}
```

which would be used for config mapping.

Given that we probably need some config in the db at some point, it might make sense to have it a

```
struct NamespaceConfig {
  ID uuid.UUID
  ConfigID int // can be empty
}
```

internal/persistence/sql/relationtuples.go
aeneasr: That makes sense, I think we can remove the nid then from here

internal/relationtuple/manager_requirements.go
aeneasr: Could there be a case where you remove a relation, later recreate that relation with the same subject ID (but it refers to someone else / new), and now that new subject ID has access to something it shouldn't have (because the old ID had access to it but was not cleared properly)?

aeneasr:
> I can't change the migration ID for already applied migrations right? That would break existing installations as far as I can tell.

Ah, I missed that the files were moved, not created.

> I will add indexes required for the check and expand APIs, but not for user-defined queries.

SGTM

internal/persistence/sql/relationtuples.go
aeneasr: Ah I see - I forgot about duplication :)

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.mysql.up.sql
zepatrik: But the namespace ID only has to be unique within a network. Also it has to be provided by some outside source (config.yml). We only have to increment it for every project, which is a searchable scope. UUIDs on the other hand will be annoying to manage manually because it is hard to compare them and a multi step process to generate new ones manually.
Think of the namespace rather as a directory or bucket of tuples that helps you organize it. Also they will contain the subject set rewrites and effect of relations.

internal/relationtuple/manager_requirements.go
zepatrik: Yes, but that is why it is so important to have unique IDs everywhere.

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.mysql.up.sql
aeneasr: I get it, you can define the namespace with an ID in your config file:

```
namespaces:
  - id: 1
    name: foo
```

but in Ory Keto it would translate into a new row in table `namespace_configs`:

```
INSERT INTO namespace_configs VALUES (id=newUUID, config_id=1)
```

So you could use UUIDs in the relation tuples, but don't have to lock the row with high isolation levels to find the largest ID in the search space, which would be a filter + order query on a insanely large distributed database table.

internal/relationtuple/manager_requirements.go
aeneasr: Can't we design the system to be resilient against something like that as opposed to hoping that people apply it correctly?

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.mysql.up.sql
zepatrik: I think I now got what you initially meant here

> I don't know how that could be solved in the user config. Potentially with some additional table / mapping:
> 
> ```
> struct NamespaceMapping {
>   ConfigKey int
>   ID uuid.UUID
> }
> ```
>
> which would be used for config mapping.

Sounds good, so this is only used when loading namespace configs from the config file, but not e.g. from a database right? When loading from the database you directly specify the UUIDs.

internal/relationtuple/manager_requirements.go
zepatrik: We can, but that requires us to define the subject IDs and it would remove a lot of the flexibility we have with the relation tuple approach.
I would rather solve this on an API level where we add a way to delete including references, see #599

aeneasr: That makes sense, also as per our discussion in the office

internal/relationtuple/manager_isolation.go
aeneasr: This tests namespace isolation, but not network isolation, right?

internal/persistence/sql/full_test.go
```diff
@@ -1,67 +1,70 @@
...
+			})
+
+			t.Run("relationtuple.IsolationTest", func(t *testing.T) {
+				var nspaces []*namespace.Namespace
+				p0, r, _ := setup(t, dsn)
+				p1, err := sql.NewPersister(r, uuid.Must(uuid.NewV4()))
+				require.NoError(t, err)
+
+				// same registry, but different persisters only differing in the network ID
+				relationtuple.IsolationTest(t, p0, p1, addNamespace(r, nspaces))
```
zepatrik, line 67: @aeneasr this is where we call the isolation test
We have the same registry and therefore same DB connection, but initialize the persisters with different network IDs.

internal/persistence/sql/namespaces.go
```diff
@@ -0,0 +1,87 @@
+package sql
```
zepatrik, line 1: This file contains everything relevant for the mapping between the serial namespace ID and the UUID stored in the database.
Because it is expected to have few namespaces that are used all the time, I added a ristretto cache for that mapping.

internal/persistence/sql/persister.go
```diff
@@ -28,121 +29,101 @@ import (
...
-		}
+func (p *Persister) Connection(ctx context.Context) *pop.Connection {
+	return popx.GetConnection(ctx, p.conn.WithContext(ctx))
+}
 
-		if err := c.Store.(interface{ Ping() error }).Ping(); err != nil {
-			p.l.WithError(err).Error("Unable to ping the database connection, retrying.")
-			return errors.WithStack(err)
-		}
+func (p *Persister) CreateWithNetwork(ctx context.Context, v interface{}) error {
```
zepatrik, line 100: Wow this diff is super messy... Please especially look into this function here, and whether it is a good approach. IMO this is a valid use-case for `reflect`, but we should think about that.

aeneasr: So generally this looks pretty smart. Personally I feel that reflection should not belong into strict enforcement code of things that do isolation. Since we have panics here, it kind of feels ok though!

internal/driver/config/namespace_memory.go
```diff
@@ -26,14 +26,24 @@ func NewMemoryNamespaceManager(nn ...*namespace.Namespace) *memoryNamespaceManag
...
-func (s *memoryNamespaceManager) GetNamespace(_ context.Context, name string) (*namespace.Namespace, error) {
+func (s *memoryNamespaceManager) GetNamespaceByName(_ context.Context, name string) (*namespace.Namespace, error) {
 	for _, n := range *s {
 		if n.Name == name {
 			return n, nil
 		}
 	}
 
-	return nil, errors.WithStack(herodot.ErrNotFound.WithReason("unknown namespace " + name))
+	return nil, errors.WithStack(herodot.ErrNotFound.WithReason("unknown namespace with name " + name))
```
aeneasr, line 36: Reasons should start with capitals and end with punctuation and use `'` :)

```suggestion
	return nil, errors.WithStack(herodot.ErrNotFound.WithReason("Unknown namespace with name '" + name"'.))
```

internal/driver/config/namespace_memory.go
```diff
@@ -26,14 +26,24 @@ func NewMemoryNamespaceManager(nn ...*namespace.Namespace) *memoryNamespaceManag
...
+}
+
+func (s *memoryNamespaceManager) GetNamespaceByConfigID(_ context.Context, id int) (*namespace.Namespace, error) {
+	for _, n := range *s {
+		if n.ID == id {
+			return n, nil
+		}
+	}
+
+	return nil, errors.WithStack(herodot.ErrNotFound.WithReasonf("unknown namespace with id %d", id))
```
aeneasr, line 46:
```suggestion
	return nil, errors.WithStack(herodot.ErrNotFound.WithReasonf("Unknown namespace with id '%d'.", id))
```

internal/driver/config/namespace_watcher.go
```diff
@@ -168,7 +168,20 @@ func (n *NamespaceWatcher) GetNamespace(_ context.Context, name string) (*namesp
 		}
 	}
 
-	return nil, errors.WithStack(herodot.ErrNotFound.WithError("unknown namespace " + name))
+	return nil, errors.WithStack(herodot.ErrNotFound.WithError("unknown namespace with name " + name))
```
aeneasr, line 171: See comment above

internal/driver/config/namespace_watcher.go
```diff
@@ -168,7 +168,20 @@ func (n *NamespaceWatcher) GetNamespace(_ context.Context, name string) (*namesp
...
+	n.RLock()
+	defer n.RUnlock()
+
+	for _, nspace := range n.namespaces {
+		if nspace.namespace.ID == id {
+			return nspace.namespace, nil
+		}
+	}
+
+	return nil, errors.WithStack(herodot.ErrNotFound.WithErrorf("unknown namespace with id %d", id))
```
aeneasr, line 184: See comment above

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.mysql.up.sql
```diff
@@ -0,0 +1,36 @@
+-- we can not fizz this migration because there is no check constraint support in fizz
+-- and it gets cumbersome with sqlite; having a working SQL version is actually way easier
+CREATE TABLE keto_relation_tuples
+(
+    shard_id                 char(36)    NOT NULL,
+    nid                      char(36)    NOT NULL,
+    namespace_id             char(36)    NOT NULL,
```
aeneasr, line 7: This is missing a FK relation to namespace

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.up.sql
```diff
@@ -0,0 +1,49 @@
+-- we can not fizz this migration because there is no check constraint support in fizz
+-- and it gets cumbersome with sqlite; having a working SQL version is actually way easier
+CREATE TABLE keto_relation_tuples
+(
+    shard_id                 UUID        NOT NULL,
+    nid                      UUID        NOT NULL,
+    namespace_id             UUID        NOT NULL,
```
aeneasr, line 7: This is missing a FK relation to namespace

internal/persistence/sql/migrations/templates/20210623162418_namespace_id.up.fizz
```diff
@@ -0,0 +1,9 @@
+create_table("keto_namespace_ids") {
+    t.Column("nid", "uuid", {})
+    t.Column("id", "uuid", {})
+    t.Column("serial_id", "integer", {})
+
+    t.PrimaryKey("nid", "id")
+}
+
+add_index("keto_namespace_ids", ["nid", "serial_id"], {unique: true})
```
aeneasr, line 9: Are we looking up by serial_id? If so please add an index

internal/persistence/sql/persister.go
```diff
@@ -28,121 +29,101 @@ import (
...
-	var opts []instrumentedsql.Opt
-	if p.tracer.IsLoaded() {
-		opts = []instrumentedsql.Opt{
-			instrumentedsql.WithTracer(opentracing.NewTracer(true)),
-			instrumentedsql.WithOmitArgs(),
-		}
+	namespaceIDCache, err := ristretto.NewCache(&ristretto.Config{
+		// TODO probably make this configurable
+		NumCounters: 10 * 1000,
+		MaxCost:     1000,
```
aeneasr, line 75: This will hold 1000 cost units - the num counters is correct (but use a const and just do 10*const for numcounters).

Since it appears that you want to store 1000 namespace objects, you will need to disable the internal cost calculator so that 1 namespace really has 1 space. Otherwise the key is included in the cost in ristrettoy:

```suggestion
		MaxCost:     1000,
		IgnoreInternalCost: true,
```

internal/persistence/sql/namespaces.go
```diff
@@ -0,0 +1,87 @@
...
+	UpdatedAt time.Time `db:"updated_at"`
+}
+
+func (*namespaceRecord) TableName(context.Context) string {
+	return "keto_namespace_ids"
+}
+
+func (p *Persister) cacheNamespaceMapping(serialID int, id uuid.UUID) {
+	p.namespaceIDCache.Set(serialID, id, 1)
+	p.namespaceIDCache.Set(id[:], serialID, 1)
```
aeneasr, line 28: Either prefix the caches with e.g. `serial-${id}`, `${pk}-id` or use two separate caches to avoid accidentally confusing these values.

internal/persistence/sql/namespaces.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return uuid.Nil, err
+	}
+
+	if id, ok := p.namespaceIDCache.Get(n.ID); ok {
+		return id.(uuid.UUID), nil
+	}
+
+	dbEntry := namespaceRecord{}
+	if err := p.QueryWithNetwork(ctx).Where("serial_id = ?", n.ID).First(&dbEntry); errors.Is(err, sql.ErrNoRows) {
```
aeneasr, line 49: `QueryWithNetwork` big brain

aeneasr (changes requested):

internal/relationtuple/manager_isolation.go
```diff
@@ -0,0 +1,75 @@
...
+			require.NoError(t, m0.DeleteRelationTuples(ctx, rts...))
+
+			deleted, _, err := m0.GetRelationTuples(ctx, &RelationQuery{Namespace: nspace})
+			require.NoError(t, err)
+			assert.Len(t, deleted, 0)
+
+			actual, _, err := m1.GetRelationTuples(ctx, &RelationQuery{Namespace: nspace})
+			require.NoError(t, err)
+			assert.Equal(t, rts, actual)
+		})
```
aeneasr, line 73: I know that `TransactRelationTuples` uses other internals, but to ensure that all public methods of the persister are tested for isolation, please still add one for `TransactRelationTuples`. Maybe the method changes later on and someone forgets to impelment isolation!

internal/persistence/sql/namespaces.go
```diff
@@ -0,0 +1,87 @@
...
+	SerialID  int       `db:"serial_id"`
+	CreatedAt time.Time `db:"created_at"`
+	UpdatedAt time.Time `db:"updated_at"`
+}
+
+func (*namespaceRecord) TableName(context.Context) string {
+	return "keto_namespace_ids"
+}
+
+func (p *Persister) cacheNamespaceMapping(serialID int, id uuid.UUID) {
```
aeneasr, line 26: Add test :)

internal/persistence/sql/namespaces.go
```diff
@@ -0,0 +1,87 @@
...
+	p.namespaceIDCache.Set(serialID, id, 1)
+	p.namespaceIDCache.Set(id[:], serialID, 1)
+}
+
+// GetNamespaceID returns the internal UUID for use in the relation tuple table.
+// This additional indirection is to allow distributed configs later on, so that we don't have
+// to sync on serial configIDs by directly using UUIDs.
+func (p *Persister) GetNamespaceID(ctx context.Context, name string) (uuid.UUID, error) {
```
aeneasr, line 34: Add test :)

Include isolation tests too please!

internal/persistence/sql/namespaces.go
```diff
@@ -0,0 +1,87 @@
...
+
+	p.cacheNamespaceMapping(n.ID, dbEntry.ID)
+
+	return dbEntry.ID, nil
+}
+
+// GetNamespaceConfigID returns the serial ID associated with the namespace
+// as stored in the relation tuple table.
+// Reverse lookup of GetNamespaceID.
+func (p *Persister) GetNamespaceConfigID(ctx context.Context, id uuid.UUID) (*namespace.Namespace, error) {
```
aeneasr, line 69: Add test :)

Include isolation tests too please!

aeneasr (changes requested):

internal/persistence/sql/migrations/templates/20210623162418_namespace_id.up.fizz
zepatrik: This is already the index for that, as we always only look within one network ID.

internal/persistence/sql/namespaces.go
zepatrik: Isn't this already handled by the different types the keys have?

internal/persistence/sql/migrations/templates/20210623162417_relationtuple.mysql.up.sql
zepatrik: good catch

internal/persistence/sql/namespaces.go
aeneasr: Yeah, but I would prefer to make it explicit. Cache key confusion is a security risk here.

zepatrik: The cache is currently shared by all networks so need to fix that in the first place 😅

internal/relationtuple/manager_isolation.go
aeneasr: ^ @zepatrik

aeneasr: https://github.com/ory/keto/pull/638#discussion_r669556863

PR #671, opened 2021-08-02
chore: adjust CODEOWNERS

zepatrik:


.github/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+*            @zepatrik
+
+/docs/docs/  @ory/documenters
```
aeneasr, line 3: I think we could even give `/docs` as they could merge also changes to the template plus versioned docs :)

aeneasr (commented):
Nice! Could you also add this to all the other repos?

aeneasr:
I mean just the big repos, not every one of the 150

.github/CODEOWNERS
zepatrik: ah right forgot versioned docs

zepatrik:
worked :tada: 
Screenshot from 2021-08-02 14-13-00

zepatrik:
> I mean just the big repos, not every one of the 150

sure, on my way

PR #707, opened 2021-09-13
feat: migration to single table SQL schema

### LINKED ISSUE #628: Migration path for v0.7

zepatrik:
## Related issue(s)

closes #628 

## Further Comments

There is a new GH action now that runs an "e2e" kind of script. It first downloads Keto v0.6.0-alpha.3, applies it's migrations, and then adds some data. After that we migrate the data over with the current checked out branch and then use the check-API to see if the data are available and work properly. The edge cases are covered in the unit tests.

---

## Commit Message (WIP)

This change adds a migration path from Keto version v0.6.x to the new persistence structure introduced by https://github.com/ory/keto/pull/638. Every namespace has to be migrated separately, or you can use the CLI to detect and migrate all namespaces at once. Have a look at `keto help namespace migrate legacy` for all details.
**Please make sure that you backup the database before running the migration command**. For live-migrations, it is possible to run Keto v0.6 along v0.7 while migrating. Please note that this migration might be a bit slower than usual, as we have to pull the data from the database, transcode it in Keto, and then write it to the new table structure.
Future versions of Keto will not include this migration script, so you will have to migrate to v0.7 to move on from there.


internal/persistence/sql/migrations/sql/20210914134624000000_legacy-cleanup.cockroach.up.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE "keto_namespace";
```
zepatrik, line 1: This table was actually never used, not even in the old versions of Keto. Therefore just dropping it.

cmd/namespace/migrate_legacy.go
```diff
@@ -0,0 +1,104 @@
...
+
+			for _, n := range nn {
+				if err := migrator.MigrateNamespace(cmd.Context(), n); err != nil {
+					_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "Encountered error while migrating: %s\nAborting.\n", err.Error())
+					return cmdx.FailSilently(cmd)
+				}
+				if err := migrator.MigrateDown(cmd.Context(), n); err != nil {
+					_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "Could not migrate down: %s\n", err.Error())
+					return cmdx.FailSilently(cmd)
+				}
```
zepatrik, line 93: I was first unsure whether we should migrate down directly without user consent, but I think it is fine because we do it after we have successfully transferred the data. Otherwise we would need a separate command just to migrate the namespace down, which is a bit of an overkill IMO.

cmd/migrate/up.go
```diff
@@ -16,18 +16,22 @@ import (
...
 
 func newUpCmd() *cobra.Command {
-	var allNamespaces bool
-
 	cmd := &cobra.Command{
 		Use:   "up",
 		Short: "Migrate the database up",
-		Long: "Migrate the database up.\n" +
-			"This does not affect namespaces. Use `keto namespace migrate up` for migrating namespaces.",
+		Long: `Run this command on a fresh SQL installation and when you upgrade Ory Keto to a new minor version.
```
aeneasr, line 26:
```suggestion
		Long: `Run this command on a fresh SQL installation and when you upgrade Ory Keto from version v0.7.0 and later.
```

cmd/namespace/migrate_legacy.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/ory/keto/internal/driver"
+	"github.com/ory/keto/internal/namespace"
+	"github.com/ory/keto/internal/persistence"
+	"github.com/ory/keto/internal/persistence/sql/migrations"
+)
+
+func NewMigrateLegacyCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "legacy [<namespace-name>]",
+		Short: "Migrate a namespace from a legacy table.",
```
aeneasr, line 21:
```suggestion
		Short: "Migrate a namespace from v0.6.x to v0.7.x and later.",
```

cmd/namespace/migrate_legacy.go
```diff
@@ -0,0 +1,104 @@
...
+		Long: "Migrate a legacy (v0.6.0) table to the v0.7.0 table.\n" +
```
aeneasr, line 22:
```suggestion
		Long: "Migrate a legacy namespaces from v0.6.x to the v0.7.x and later.\n" +
```

cmd/namespace/migrate_legacy.go
aeneasr: Hm, the problem here would be that if something goes wrong, the old data is gone. We should definitely respect this when executing this on our DB!

internal/driver/pop_connection.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/cenkalti/backoff/v3"
+	"github.com/gobuffalo/pop/v5"
+	"github.com/luna-duclos/instrumentedsql"
+	"github.com/luna-duclos/instrumentedsql/opentracing"
+	"github.com/ory/x/sqlcon"
+	"github.com/pkg/errors"
+)
+
+func (r *RegistryDefault) PopConnectionWithOpts(popOpts ...func(*pop.ConnectionDetails)) (*pop.Connection, error) {
```
aeneasr, line 14: Looks like something we could ory/x at some point :)

scripts/single-table-migration-e2e.sh
```diff
@@ -0,0 +1,38 @@
...
+
+./keto serve all -c ./contrib/cat-videos-example/keto.yml &
+keto_server_pid=$!
+
+function teardown() {
+    kill $keto_server_pid || true
+}
+trap teardown EXIT
+
+jq '[range(200)] | map({namespace: "videos", object: . | tostring, relation: "view", subject: "user"})' <(echo '{}') \
```
aeneasr, line 23: Should we create objects in multiple namespaces to ensure that all namespaces are migrated?

internal/persistence/sql/migrations/single_table.go
```diff
@@ -0,0 +1,271 @@
...
+		if err := p.Transaction(ctx, func(ctx context.Context, _ *pop.Connection) error {
+			rs, hasNext, err := m.getOldRelationTuples(ctx, n, page, m.perPage)
+			if err != nil {
+				return err
+			}
+
+			for _, r := range rs {
+				ri, err := r.toInternal()
+				if err != nil {
+					m.d.Logger().WithField("relation_tuple", r).WithField("hint", "https://github.com/ory/keto/issues/661").WithError(err).Warn("Skipping relation tuple, it seems to be in a broken state. Please recreate it manually.")
```
aeneasr, line 186: Can this actually happen? How would one recover from this, given that we delete the data after the migration is done?

internal/persistence/sql/migrations/single_table.go
```diff
@@ -0,0 +1,271 @@
...
+func (m *toSingleTableMigrator) insertOldRelationTuples(ctx context.Context, n *namespace.Namespace, rs ...*relationtuple.InternalRelationTuple) error {
+	for _, r := range rs {
+		if r.Subject == nil {
+			return errors.New("subject is not allowed to be nil")
+		}
+
+		m.d.Logger().WithFields(r.ToLoggerFields()).Trace("creating in legacy database")
+
+		if err := m.d.Persister().Connection(context.WithValue(ctx, namespaceCtxKey, n)).Create(&relationTuple{
+			ID:         "testing only",
```
aeneasr, line 158: What implications does this have?

internal/persistence/sql/migrations/single_table_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"github.com/ory/keto/internal/driver"
+	"github.com/ory/keto/internal/driver/config"
+	"github.com/ory/keto/internal/namespace"
+	"github.com/ory/keto/internal/relationtuple"
+	"github.com/ory/keto/internal/x"
+	"github.com/ory/keto/internal/x/dbx"
+)
+
+func TestToSingleTableMigrator(t *testing.T) {
```
aeneasr, line 19: Am I correct in assuming that we are testing a single namespace here also? Would it make sense to ensure that we properly detect all namespaces if we have multiple namespaces?

aeneasr (commented):
Looking great! :)

internal/persistence/sql/migrations/single_table.go
zepatrik: As described in the issue, you can theoretically create a tuple that is serialized for the legacy table into a format that is not de-serializable anymore. The only way to recover is to recreate the tuple once the rest is migrated. But you are right, we should maybe not migrate down in this case.

cmd/namespace/migrate_legacy.go
zepatrik: But if something goes wrong before, we would have already returned.

internal/persistence/sql/migrations/single_table.go
zepatrik: This function is only used for testing, because we first have to insert some tuples before we can migrate them over.
`ID` is the `shard_id` column.

internal/persistence/sql/migrations/single_table_test.go
zepatrik: The function can only migrate a single namespace. We create a new one for each subtest.
The test [below](https://github.com/ory/keto/pull/707/files#diff-d8f4c5835169af823f074a4f9c32416e4bbc979b24a0c37ad1dfdbee095e7ec1R129) does the namespace detection, but only one. Will add multiple ones there.

scripts/single-table-migration-e2e.sh
zepatrik: True :+1:

internal/persistence/sql/migrations/single_table_test.go
aeneasr: Nice, ty!

cmd/namespace/migrate_legacy.go
aeneasr: I assume the comment is no longer relevant as per https://github.com/ory/keto/pull/707/files#r712027278

scripts/single-table-migration-e2e.sh
aeneasr: Is this already resolved? If not please resolve after they have been resolved in code :)

zepatrik: I have it  resolved but not yet pushed :sweat_smile: CCI was too busy already yesterday for intermediate commits.

aeneasr (approved):

PR #713, opened 2021-09-15
ci: fix go gRPC release script

zepatrik:
fixes stuff from #712


zepatrik (commented):
Seems to work fine now: https://github.com/ory/keto/releases/tag/proto%2Fv0.6.0-alpha.3-testrelease

PR #730, opened 2021-09-23
docs: change npm org scope from @oryd to @ory

swantzter:
https://www.npmjs.com/package/@oryd/keto-client has a notice that @oryd has changed to @ory, and https://www.npmjs.com/package/@oryd/keto-grpc-client simply doesn't exist (whereas the @ory version does)

### DIFF
--- a/docs/versioned_docs/version-v0.6/sdk/index.md
+++ b/docs/versioned_docs/version-v0.6/sdk/index.md
@@ -28,9 +28,9 @@ repositories:
 - [Python](https://pypi.org/project/ory-keto-client/)
 - [Ruby](https://rubygems.org/gems/ory-keto-client)
 - [Rust](https://crates.io/crates/ory-keto-client)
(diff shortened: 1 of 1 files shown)


aeneasr:
Thank you very much for your contribution to the documentation! Due to the way [Docusaurus](https://docusaurus.io) versioning works, you need to make the changes in two locations:

- `docs/versioned_docs/version-v{version}/path/to/file`
- `docs/docs/path/to/file`

Without making them in both locations, the changes will most likely be lost when releasing another version. Thank you! :)

swantzter:
@aeneasr it looks like the change has already been made in `docs/docs` and are only outdated in the versioned docs, would you like me to update in v0.5 docs too? Though that only points to https://www.npmjs.com/package/@oryd/keto-client which does exist but is deprecated, yet it's still the correct version for v0.5 from what I can see

zepatrik (approved):
Makes sense in this case. The new package does not have the old versions, so this is good :+1:

PR #755, opened 2021-10-05
Fix #704 `max-depth` not optional.

### LINKED ISSUE #704: Keto expand REST API does not parse query parameter `max-depth`
**Describe the bug**

Keto exapnd REST API will return http code 400 when there is not  `max-depth` in query parameter. 

**Reproducing the bug**
image

https://github.com/ory/keto/blob/d766968419d10a68fd843df45316e3436b68d61d/internal/expand/handler.go#L77-L82

**Expected behavior**

It should behave as [doc](https://www.ory.sh/keto/docs/reference/rest-api#expand-a-relation-tuple) says that `max-depth` is **not required**.

Or it should be consistent with cli behavior, that's to say, has a default value

```shell
$ keto expand grant role company-admin

∪ role:company-admin#grant
├─ ☘ role:super-admin️
├─ ∪ role:super-admin#member
│  ├─ ☘ user:ZhangThree️
```

**Environment**

- Version: 0.6.0-alpha.3
- Environment: Docker compose, PostMan

CypherpunkSamurai:
# Fix #704
`max-depth` not working as optional as mentioned in [docs](https://www.ory.sh/keto/docs/reference/rest-api#expand-a-relation-tuple).

## Related issue(s)

#704 

## Checklist

- [x] I am following the
      the changes.
      works.

## Further Comments

This patch fixes issue #704 where `max-depth` parameter appears to be causing http errors (HTTP 400).

This patch reads the query parameter `max-depth` into variable `d_` and checks if query parameter was not provided before throwing http error.


internal/expand/handler.go
```diff
@@ -76,8 +76,9 @@ type getExpandRequest struct {
 //       404: genericError
 //       500: genericError
 func (h *handler) getExpand(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
-	depth, err := strconv.ParseInt(r.URL.Query().Get("max-depth"), 0, 0)
-	if err != nil {
+	d_ := r.URL.Query().Get("max-depth")
```
zepatrik, line 79: With go1.17 we can now use `r.URL.Query().Has("max-depth")` https://pkg.go.dev/net/url#Values.Has :wink: 
So I would suggest something along the lines of
```suggestion
	if !r.URL.Query().Has("max-depth") {
		h.d.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError(`required parameter "max-depth" is missing`)
		return
	}
```

zepatrik (changes requested):

internal/expand/handler.go
CypherpunkSamurai: Sure. I'll add them in a sec.

zepatrik:
Further, please add a test that checks whether the right code is returned.
To actually make the parameter required, please add a `// required: true` comment here: https://github.com/ory/keto/blob/4e65c14b28daf7ab7ec0dc40c95885c9adc33958/internal/expand/handler.go#L55-L56
You will find other definitions that do that, so please look up the right way to do this.
The parameter is not marked as required in the docs (anymore?): https://www.ory.sh/keto/docs/reference/rest-api/#operation/getExpand

internal/expand/handler.go
CypherpunkSamurai: Excuse me for my novice question, but isn't `max-depth` supposed to be optional parameter?

zepatrik: Yes, sorry missed that. Thinking about it a second time, it probably makes more sense to make it actually required as I can't come up with a sane default value that makes sense for most use-cases. Currently it is required but that is neither reflected in the API spec, nor is a proper (i.e. helpful) error returned.

CypherpunkSamurai:
> Yes, sorry missed that. Thinking about it a second time, it probably makes more sense to make it actually required as I can't come up with a sane default value that makes sense for most use-cases.

As you're planning on making it required I'll make necessary changes as you suggested. This would make things more clearer :)

Also, side note, should I change the documentation files too to reflect the changes?

internal/expand/handler.go
CypherpunkSamurai: Ok. I'll make necessary changes and update the patch.

docs/docs/.static/api.json
```diff
@@ -282,7 +282,8 @@
             "type": "integer",
             "format": "int64",
             "name": "max-depth",
-            "in": "query"
+            "in": "query",
+	    "required": true
```
zepatrik, line 286: This file is auto-generated from the comments in the code. Please run `make sdk` to update it.

internal/expand/handler.go
```diff
@@ -52,7 +52,7 @@ func (h *handler) RegisterWriteGRPC(s *grpc.Server) {}
 // swagger:parameters getExpand
 // nolint:deadcode,unused
 type getExpandRequest struct {
-	// in:query
+	// required: true
```
zepatrik, line 55: You will need both of these comments, as it is in the query and is required.

internal/expand/handler.go
```diff
@@ -76,9 +76,13 @@ type getExpandRequest struct {
...
 //       500: genericError
 func (h *handler) getExpand(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	if !r.URL.Query().Has("max-depth") {
+		h.d.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError(`required parameter "max-depth" is missing`)
+		return
+	}
 	depth, err := strconv.ParseInt(r.URL.Query().Get("max-depth"), 0, 0)
 	if err != nil {
-		h.d.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError(err.Error()))
+		h.d.Writer().WriteError(w, r, err)
```
zepatrik, line 85: This should still be a `herodot.ErrBadRequest` because that will correctly set the status code and more.
The error at this point is a failure to parse an integer, so it is always a bad request.

internal/expand/handler_test.go
```diff
@@ -35,6 +35,17 @@ func TestRESTHandler(t *testing.T) {
 	ts := httptest.NewServer(r)
 	defer ts.Close()
 
+
+	t.Run("case=returns required parameter \"max-depth\" is missing", func(t *testing.T) {
```
zepatrik, line 39: Nice :+1:

zepatrik (changes requested):
Looking pretty good already :+1: 
Did you find any docs pages that need to be updated except for the API reference? That one is automatically rendered from the openapi schema, which is autogenerated from the code and comments. So should be enough to just adjust the annotations.

docs/docs/.static/api.json
CypherpunkSamurai: Ok. Sure

internal/expand/handler.go
CypherpunkSamurai: Ok. Noted

CypherpunkSamurai:
I found a new problem after adding the patch to check `max-depth`. Is the function supposed to use `max-depth` as a `*int64` rather than a `int64`?

image

internal/expand/handler_test.go
```diff
@@ -35,6 +35,16 @@ func TestRESTHandler(t *testing.T) {
...
 	defer ts.Close()
 
+	t.Run("case=returns required parameter \"max-depth\" is missing", func(t *testing.T) {
+		resp, err := ts.Client().Get(ts.URL + expand.RouteBase)
+		require.NoError(t, err)
+
+		assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
+		body, err := io.ReadAll(resp.Body)
+		require.NoError(t, err)
+		assert.Contains(t, string(body), "required parameter \"max-depth\"")
```
zepatrik, line 45: The tests are failing here now because the body is a JSON object, so the string escapes the quotes.

internal/expand/handler.go
```diff
@@ -76,6 +77,10 @@ type getExpandRequest struct {
 //       404: genericError
 //       500: genericError
 func (h *handler) getExpand(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	if !r.URL.Query().Has("max-depth") {
+		h.d.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError(`required parameter "max-depth" is missing`))
```
zepatrik, line 81: I suggest you adjust the error message here:
```suggestion
		h.d.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError("required query parameter 'max-depth' is missing"))
```
which will make it a bit more helpful and also avoid double-escaping in the test.

internal/e2e/sdk_client_test.go
```diff
@@ -213,7 +213,7 @@ func (c *sdkClient) expand(t require.TestingT, r *relationtuple.SubjectSet, dept
 			WithNamespace(r.Namespace).
 			WithObject(r.Object).
 			WithRelation(r.Relation).
-			WithMaxDepth(pointerx.Int64(int64(depth))),
+			WithMaxDepth(int64(depth)),
```
zepatrik, line 216: This change to the SDK is expected, so everything is fine here.

CypherpunkSamurai:
Ok. All checks seems to have passed. Thank You for being patient with me. I'm new to PRs.

I'm still learning please excuse my commits. I'm very sorry for taking so much time.

I had a question and I feel like you're a great developer who can answer. How does one test their code offline before making PR? Everytime I push I see the CI takes over and runs the checks for me. Is it possible to run tests offline?

PR #791, opened 2021-12-05
feat: add max-depth for check command + global max-depth

### LINKED ISSUE #702: Add `max-depth` to Check API

It would be nice to have a `max-depth` parameter to the check API so checks could not go past a specific depth (and might fail as a result)

Adding `max-depth` as a parameter

**Describe alternatives you've considered**

Expanding instead of checking... bad solution

**Additional context**

In the case of checks in a multi-tenant environment it would be great to prevent highly nested checks.

### LINKED ISSUE #703: Add `max-depth` to global config

It would be nice for multi-tenant environmentss to be able to specify a global `max-depth` such that requests would not be able to cause high latency and demanding DB queries.

Adding `max-depth` as a something the `max-depth` query param cannot exceed.

**Describe alternatives you've considered**

None

**Additional context**

doodlesbykumbi:
max-depth for the check command limits the depth of the search, a safeguard against particularly expensive queries. This allows users more fine-grain control.

At present the approach in this PR is to 
1. Add max-depth to the check CLI command, and the API (protobufs + HTTP query params)
2. When max-depth is hit for check, which is compared against the number of recursive (beyond the first) calls to `checkOneIndirectionFurther`, the search along the given branch halts. Check returning false means a path to a relation is has not been found up to the max-depth.
3. Adds a global max-depth for all read operations accessible via the`serve.read.max-depth` key in the config json. This global max-depth takes precedence if the request-level max-depth is less than 1 or if the global max-depth is less than the request-level max depth. 

## Related issue(s)

closes #703 
closes #702 

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Notes for self

1. Try to run keto locally. 
This was straightforward and so I did so via Github Codespaces. The Makefile shows me how to do most common development activities. I was able to run the cat video example with a local server.
2. Try to understand the structure of the code.
    + The app is a CLI (based on cobra) that can serve and behave as a client.
    + Serving leveraging gRPC. The version of protoc used is v3.13.0, gleamed from the auto-generated Go files.
    + The "expand" command also makes use of max-depth and so I can crib from that
    + Adding max-depth to "check" command requires similar changes to thosse in the "expand" command: these can be roughly categorised as docs, client, server
    + Running the examples, exploring the code and tests around the "expand" command sheds light on the data model
3. Implement
    + Make changes to docs and client, to expose maxDepth param and thread it all the way to the "check" handler
    + Make sense of the "check" handler. It uses recursive depth first search?! `checkOneIndirectionFurther` is the method that is called recursively. NOTE: The search for `keto check "*" view videos /cats/1.mp4` wastes diving 2 units of depth while the answer is present at 0 depth.
    + Implement first pass and confirm it through local dev and units
    + Response thoughts
        + Initial thoughts are to decrement maxDepth for each recursive call of `checkOneIndirectionFurther` and error when maxDepth is spent
        + Alternative approach is to NOT error but instead return the value of "allowed" up to the maxDepth


internal/check/handler.go
```diff
@@ -106,6 +120,20 @@ func (h *Handler) getCheck(w http.ResponseWriter, r *http.Request, _ httprouter.
 	h.d.Writer().WriteCode(w, r, http.StatusForbidden, &RESTResponse{Allowed: false})
 }
 
+func (h *Handler) ensureDepthQueryParam(w http.ResponseWriter, r *http.Request) (int, bool)  {
```
zepatrik, line 123: This helper function could also be used by the expand API right? We could refactor it to be in the `internal/x` package for example.

zepatrik: I would propose to then make it `func GetMaxDepthFromQuery(q *url.Values) error`.

proto/install.sh
```diff
@@ -0,0 +1,5 @@
+PB_REL="https://github.com/protocolbuffers/protobuf/releases"
+PB_VERION="3.13.0"
+curl -LO $PB_REL/download/v${PB_VERION}/protoc-${PB_VERION}-linux-x86_64.zip
+unzip protoc-${PB_VERION}-linux-x86_64.zip -d $HOME/.local
+export PATH="$PATH:$HOME/.local/bin"
```
zepatrik, line 5: I installed protoc from the package manager, but this should ideally be part of the Makefile. However, this script will not be sufficient for everyone as it only works on one system/arch combination. I would rather exclude it for now and add a different installation method.

internal/check/engine.go
```diff
@@ -90,6 +112,6 @@ func (e *Engine) checkOneIndirectionFurther(ctx context.Context, requested *rela
 	}
 }
 
-func (e *Engine) SubjectIsAllowed(ctx context.Context, r *relationtuple.InternalRelationTuple) (bool, error) {
-	return e.checkOneIndirectionFurther(ctx, r, &relationtuple.RelationQuery{Object: r.Object, Relation: r.Relation, Namespace: r.Namespace})
+func (e *Engine) SubjectIsAllowed(ctx context.Context, r *relationtuple.InternalRelationTuple, restDepth int) (bool, error) {
+	return e.checkOneIndirectionFurther(ctx, r, &relationtuple.RelationQuery{Object: r.Object, Relation: r.Relation, Namespace: r.Namespace}, restDepth)
```
zepatrik, line 116: At this point one can easily also add a fix for #703, just take the min of the global and per-request value here. I am adding this comment to give a hint to whoever will work on it in the end in a separate PR.

zepatrik (changes requested):
Thanks, this looks very good already :+1:
I only have a view small comments, but no hurry :wink: 

Regarding your open questions, I agree with you that returning an error is not the right way to go. Instead, only the branch should be aborted. We could at some point add a boolean to the response that indicates whether not all branches were fully expanded, but I am not sure that there is a good use-case for that.
Except maybe of debugging, but for that you can add a log statement instead of returning an error.

Also, I think the `max-depth` parameter should not be required, but instead optional. With #703 we will have a configurable fallback that will be used if no value is given.

Thanks again for looking into this.

internal/check/handler.go
doodlesbykumbi: Thanks. Addressed

proto/install.sh
doodlesbykumbi: Thanks. Addressed with a comment in the README instead

internal/check/engine.go
doodlesbykumbi: Thanks. In fact, I went ahead and added the global value feature to this PR

doodlesbykumbi:
@zepatrik this should be good for re-review

internal/check/engine.go
```diff
@@ -90,6 +111,11 @@ func (e *Engine) checkOneIndirectionFurther(ctx context.Context, requested *rela
 	}
 }
 
-func (e *Engine) SubjectIsAllowed(ctx context.Context, r *relationtuple.InternalRelationTuple) (bool, error) {
-	return e.checkOneIndirectionFurther(ctx, r, &relationtuple.RelationQuery{Object: r.Object, Relation: r.Relation, Namespace: r.Namespace})
+func (e *Engine) SubjectIsAllowed(ctx context.Context, r *relationtuple.InternalRelationTuple, restDepth int) (bool, error) {
+	// global max-depth takes precedence when it is the lesser or if the request max-depth is less than or equal to 0
+	if globalMaxDepth := e.d.Config().ReadAPIMaxDepth(); restDepth <= 0 || globalMaxDepth < restDepth {
```
zepatrik, line 116: I see, so `0` means fallback to global.

internal/check/handler.go
```diff
@@ -63,6 +63,14 @@ type RESTResponse struct {
 	Allowed bool `json:"allowed"`
 }
 
+// swagger:parameters getCheckRequest
+// nolint:deadcode,unused
+type getCheckRequest struct {
+	// in:query
+	// required: true
```
zepatrik, line 70:
```suggestion
```
It is not required anymore, right?

internal/check/handler.go
```diff
@@ -63,6 +63,14 @@ type RESTResponse struct {
 	Allowed bool `json:"allowed"`
 }
 
+// swagger:parameters getCheckRequest
```
zepatrik, line 66: You have to add the post route here as well for the SDK.
```suggestion
// swagger:parameters getCheckRequest postCheck
```

internal/check/handler.go
```diff
@@ -126,12 +140,19 @@ func (h *Handler) getCheck(w http.ResponseWriter, r *http.Request, _ httprouter.
 //       403: getCheckResponse
 //       500: genericError
 func (h *Handler) postCheck(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	maxDepth, err := x.GetMaxDepthFromQuery(r.URL.Query(), false)
```
zepatrik, line 143: Hm, just thinking about whether this endpoint, as it reads all other values from the JSON body, should also consume the `max-depth` from the body. What do you think about this inconsistency? On the other hand, `max-depth` is a kind of meta-parameter, it will never contain sensitive information.

internal/check/handler.go
```diff
@@ -126,12 +140,19 @@ func (h *Handler) getCheck(w http.ResponseWriter, r *http.Request, _ httprouter.
...
+	if err != nil {
+		h.d.Writer().WriteError(w, r, herodot.ErrBadRequest.WithError(err.Error()))
+		return
+	}
+
 	var tuple relationtuple.InternalRelationTuple
 	if err := json.NewDecoder(r.Body).Decode(&tuple); err != nil {
 		h.d.Writer().WriteError(w, r, errors.WithStack(herodot.ErrBadRequest.WithReasonf("Unable to decode JSON payload: %s", err)))
+		return
```
zepatrik, line 152: Haha, spotted that one also yesterday in the review :wink:

internal/expand/engine.go
```diff
@@ -28,8 +28,9 @@ func NewEngine(d EngineDependencies) *Engine {
 }
 
 func (e *Engine) BuildTree(ctx context.Context, subject relationtuple.Subject, restDepth int) (*Tree, error) {
-	if restDepth <= 0 {
-		return nil, nil
+	// global max-depth takes precedence when it is the lesser or if the request max-depth is less than or equal to 0
+	if globalMaxDepth := e.d.Config().ReadAPIMaxDepth(); restDepth <= 0 || globalMaxDepth < restDepth {
```
zepatrik, line 32: Nice catch here :+1:

internal/check/engine.go
```diff
@@ -66,7 +78,16 @@ func (e *Engine) subjectIsAllowed(ctx context.Context, requested *relationtuple.
...
 
-func (e *Engine) checkOneIndirectionFurther(ctx context.Context, requested *relationtuple.InternalRelationTuple, expandQuery *relationtuple.RelationQuery) (bool, error) {
+func (e *Engine) checkOneIndirectionFurther(
+	ctx context.Context,
+	requested *relationtuple.InternalRelationTuple,
+	expandQuery *relationtuple.RelationQuery,
+	restDepth int,
+) (bool, error) {
+	if restDepth <= 0 {
+		return false, nil
```
zepatrik, line 88: For debugging purposes, it would be helpful to log here that the limit was reached.
```suggestion
		e.d.Logger().WithFields((*relationtuple.InternalRelationTuple)(expandQuery).ToLoggerFields()).Debug("reached max-depth, therefore this query will not be further expanded")
		return false, nil
```

internal/x/max_depth.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"fmt"
+	"net/url"
+	"strconv"
+)
+
+func GetMaxDepthFromQuery(q url.Values, required bool) (int, error) {
+	if !q.Has("max-depth") {
+		if required {
+			return 0, fmt.Errorf("required query parameter 'max-depth' is missing")
```
zepatrik, line 12: We usually return `errors.WithStack` as a convention. Also, you could make these already `herodot.ErrBadRequest.WithError()`, so that the response status codes are set properly. All errors are bad requests, so it does not matter too much, but it would allow the handlers to directly write the error. In case there will be a new case at some point, it would also allow the use of different status codes.

proto/ory/keto/acl/v1alpha1/check_service.proto
```diff
@@ -73,6 +73,8 @@ message CheckRequest {
   // ACLs had already been replicated to all availability zones.
   // -->
   string snaptoken = 6;
+  // The maximum depth to search for a relation.
```
zepatrik, line 76: It would be nice to add a short text about the default value here.

zepatrik (changes requested):
Nice, thanks for all the improvements :+1: 
There seem to be some not cleaned up test files, I am not sure why sqlite is sometimes keeping them... 
- internal/persistence/sql/TestRelationTupleSubjectTypeCheck
- internal/persistence/sql/migrations/TestToSingleTableMigrator_HasLegacyTable
- internal/persistence/sql/migrations/migratest/TestMigrations

internal/check/handler.go
doodlesbykumbi: Yeah, and I suppose it's not required for expand either

internal/check/engine.go
doodlesbykumbi: Yeah, I tried to follow the logic I found in the codebase. Instead of validation and errors there are safe defaults or sane response. Maybe some follow up work could be more explicit in validating and providing detailed feedback to the user

zepatrik: Sounds good to me :+1:

internal/check/handler.go
doodlesbykumbi: I thought to put the parameter in the JSON body initially but I think a query param is fine. It definitely is a meta-parameter, and perhaps the right place for those is in the query params.

internal/check/engine.go
doodlesbykumbi: Type conversion won't work here since `RelationQuery` isn't a subset of `InternalRelationTuple`

internal/x/max_depth.go
doodlesbykumbi: For this I thought `GetMaxDepthFromQuery` might be independent of considerations about the form the response takes, this would allow the caller to make those decisions.

proto/ory/keto/acl/v1alpha1/check_service.proto
doodlesbykumbi: Agreed. Adding

internal/relationtuple/definitions.go
```diff
@@ -660,6 +662,10 @@ func NewManagerWrapper(_ *testing.T, reg ManagerProvider, options ...x.Paginatio
 	}
 }
 
+func (t *ManagerWrapper) Config() *config.Config {
```
doodlesbykumbi, line 665: @zepatrik I realised that this was a lazy implementation because I wasn't sure how to separate ManagerWrapper from the EngineDependencies for check and expand in the tests. I think because there was a single dependency the tests have some coupling between EngineDependencies and ManagerProvider. This approach results in coupling of unrelated things. Manager and Config. Adding a Logger would exacerbate things. I'll push a commit that separates them in the tests and in production.

internal/x/max_depth.go
zepatrik: My bad, it is not anymore possible. Forgot that.

cmd/check/root.go
```diff
@@ -56,6 +64,7 @@ func newCheckCmd() *cobra.Command {
 
 	client.RegisterRemoteURLFlags(cmd.Flags())
 	cmdx.RegisterFormatFlags(cmd.Flags())
+	cmd.Flags().Int32P(FlagMaxDepth, "d", 0, "Maximum depth of the search tree. If the value is less than 1 or greater than the global max-depth then the global max-depth will be used instead.")
```
zepatrik, line 67: I like this :+1:

internal/check/engine_test.go
```diff
@@ -20,13 +20,103 @@ import (
...
 
-func newDepsProvider(t *testing.T, namespaces []*namespace.Namespace, pageOpts ...x.PaginationOptionSetter) *relationtuple.ManagerWrapper {
+type configProvider = config.Provider
+type loggerProvider = x.LoggerProvider
+// deps is defined to capture engine dependencies in a single struct
+type deps struct {
+	*relationtuple.ManagerWrapper // managerProvider
+	configProvider
+	loggerProvider
+}
```
zepatrik, line 30: Ah, now I get what you mean. I think the previous approach was fine as well, as the manager wrapper type is only a helper for testing, but this is of course the best way to do it :+1:

zepatrik (approved):
Thank you very much, this looks perfect now :+1:

zepatrik:
Ah, the only missing thing now is the CLA :sweat_smile:

doodlesbykumbi:
@zepatrik CLA done

zepatrik:
Thank you, CCI is somehow not running... I am looking into it.

doodlesbykumbi:
@zepatrik It looks like the commit `chore: use npm@7 to generate lock-file v2` borked something in linting. You want me to look into it or should I leave that one to you ?

zepatrik:
Yeah, for some reason the lock file resulted in a broken state... should be fixed now though

zepatrik:
Thank you very much, greatly appreciate your effort :+1:

PR #799, opened 2021-12-18
feat: bulk deletion of relation tuples

### LINKED ISSUE #599: Bulk deletion of relation tuples

vancanhuit:
## Related issue(s)
Closes #599

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


internal/relationtuple/transact_server.go
```diff
@@ -170,15 +167,28 @@ func (h *handler) createRelation(w http.ResponseWriter, r *http.Request, _ httpr
...
 
-	if err := h.d.RelationTupleManager().DeleteRelationTuples(r.Context(), rel); err != nil {
-		h.d.Logger().WithError(err).WithFields(rel.ToLoggerFields()).Errorf("got an error while deleting the relation tuple")
+	l := h.d.Logger()
+	for k := range q {
+		l = l.WithField(k, q.Get(k))
+	}
+	l.Debug("deleting relation tuples")
+
+	rels, _, err := h.d.RelationTupleManager().GetRelationTuples(r.Context(), query)
```
zepatrik, line 184: This is quite problematic as you only delete the first page here. Can you instead implement the selection of delete candidates on the persistence layer? A request DELETE `/relation-tuples?namespace=foo` should result in the following SQL statement to be executed: `DELETE FROM keto_relation_tuples WHERE namespace=foo`.

zepatrik (changes requested):
This is already a good start, but needs some improvements. Also, the same behavior should be reflected over gRPC as well.

internal/relationtuple/transact_server.go
vancanhuit: Aha, yeah I will try. For the gRPC part, I may need time to understand it because I've never worked with gRPC before.

zepatrik: Ah so there is no specific delete RPC, but only the transact one. I would add a `DeleteReleationTuples` RPC here https://github.com/ory/keto/blob/67c4c608a9bbdb866701e7f3746c1d13f2e6f2f0/proto/ory/keto/acl/v1alpha1/write_service.proto#L14-L20
It should behave exactly as the REST handler, so you can probably use the same internal helper function. The current transaction handler is implemented here: https://github.com/ory/keto/blob/67c4c608a9bbdb866701e7f3746c1d13f2e6f2f0/internal/relationtuple/transact_server.go#L30-L53, you can just follow the implementation of that as a guidance.

internal/persistence/sql/relationtuples.go
```diff
@@ -200,6 +200,42 @@ func (p *Persister) DeleteRelationTuples(ctx context.Context, rs ...*relationtup
...
+			args = append(args, n.ID)
+		}
+		if query.Object != "" {
+			sqlRawQuery += " AND object = ?"
+			args = append(args, query.Object)
+		}
+		if query.Relation != "" {
+			sqlRawQuery += " AND relation = ?"
+			args = append(args, query.Relation)
+		}
```
zepatrik, line 222: Please use the pop query builder for this instead of raw query. Just as we do here: https://github.com/ory/keto/blob/7ee65b5cd0827a3de87be3b6a7fa62d9038ea2df/internal/persistence/sql/relationtuples.go#L218-L236
and here https://github.com/ory/keto/blob/7ee65b5cd0827a3de87be3b6a7fa62d9038ea2df/internal/persistence/sql/relationtuples.go#L186-L196

As is, there is no SQL injection possible, but it is easy to add one later if you don't pay attention.

zepatrik (changes requested):
Getting there :wink:

internal/persistence/sql/relationtuples.go
vancanhuit: Ah, my bad. I actually used pop query builder but for some reason, it didn't work. Now I realized why :zany_face:
I will push a new commit.

vancanhuit:
For the gRPC part, I have a basic understanding of it, trying to figure out how to do it.

internal/relationtuple/transact_server.go
```diff
@@ -52,6 +52,23 @@ func (h *handler) TransactRelationTuples(ctx context.Context, req *acl.TransactR
 	}, nil
 }
 
+func (h *handler) DeleteRelationTuples(ctx context.Context, req *acl.DeleteRelationTuplesRequest) (*acl.DeleteRelationTuplesResponse, error) {
+	if req.Query == nil {
+		return nil, errors.New("invalid request")
```
zepatrik, line 57: Always return `errors.WithStack(herodot.ErrInvalidRequest.WithReason(...))` or similar errors, they will automatically set the right response codes through the middle-ware.

internal/persistence/sql/relationtuples.go
```diff
@@ -200,6 +200,33 @@ func (p *Persister) DeleteRelationTuples(ctx context.Context, rs ...*relationtup
...
+			sqlQuery.Where("object = ?", query.Object)
+		}
+		if query.Relation != "" {
+			sqlQuery.Where("relation = ?", query.Relation)
+		}
+		if s := query.Subject(); s != nil {
+			if err := p.whereSubject(ctx, sqlQuery, s); err != nil {
+				return err
+			}
+		}
```
zepatrik, line 223: Now this looks way better, but it is the same code as in the `GetRelationTuples` function.
Can you instead add a helper function similar to this (naming is just a draft):

```go
// whereQuery takes a relation tuple query and appends where clauses to the base pop query
func whereQuery(ctx context.Context, baseQuery *pop.Query, rq *relationtuple.RelationQuery) (*pop.Query, error) {
	if query.Namespace != "" {
		n, err := p.GetNamespaceByName(ctx, rq.Namespace)
		if err != nil {
			return nil, err
		}
		baseQuery.Where("namespace_id = ?", n.ID)
	}
	if rq.Object != "" {
		baseQuery.Where("object = ?", rq.Object)
	}
	if rq.Relation != "" {
		baseQuery.Where("relation = ?", rq.Relation)
	}
	if s := rq.Subject(); s != nil {
		if err := p.whereSubject(ctx, baseQuery, s); err != nil {
			return err
		}
	}
}
```

zepatrik: And then of course use it in both functions.

internal/relationtuple/transact_server.go
```diff
@@ -52,6 +52,23 @@ func (h *handler) TransactRelationTuples(ctx context.Context, req *acl.TransactR
...
+	}
+
+	q, err := (&RelationQuery{}).FromProto((*acl.ListRelationTuplesRequest_Query)(req.Query))
```
zepatrik, line 60: The generated proto definitions have getter functions, so you can instead rewrite `func (q *RelationQuery) FromProto(...)` to take an interface with the getter functions. That way you don't have to convert the type which might break in the future at some point.

zepatrik (commented):
Now this looks very good now. The very last thing would be to add e2e tests.

- here are all cases, add some more that test the new feature: https://github.com/ory/keto/blob/7ee65b5cd0827a3de87be3b6a7fa62d9038ea2df/internal/e2e/cases_test.go
- here is the interface definition for the clients, add a `deleteAll` function and implement it in all clients: https://github.com/ory/keto/blob/7ee65b5cd0827a3de87be3b6a7fa62d9038ea2df/internal/e2e/full_suit_test.go#L33-L41

internal/persistence/sql/relationtuples.go
vancanhuit: Yeah, I will update it.

vancanhuit:
I will add e2e tests later on.

vancanhuit:
Hi, currently I'm getting stuck in the CLI client test. The `keto relation-tuple delete` command only supports JSON files containing a list of tuples. It is impossible to create a test for it based on a relation query. What should I do?

zepatrik:
I think it would make sense to not break `keto relation-tuple delete` behavior, but maybe add a flag `--all-matching` and then flags for each key you could delete by? Similar to how the list command works: https://github.com/ory/keto/blob/fabf1a06d90139f6678fc7bc1703ee92a5fc9d69/cmd/relationtuple/get.go#L22-L65

vancanhuit:
> I think it would make sense to not break `keto relation-tuple delete` behavior, but maybe add a flag `--all-matching` and then flags for each key you could delete by? Similar to how the list command works:
> 
> https://github.com/ory/keto/blob/fabf1a06d90139f6678fc7bc1703ee92a5fc9d69/cmd/relationtuple/get.go#L22-L65

Does this mean that we will support two cases: one is deleting from JSON files and one is deleting by specifying query keys?

vancanhuit:
@zepatrik How do you think about my latest change?

internal/relationtuple/transact_server.go
```diff
@@ -152,11 +166,11 @@ func (h *handler) createRelation(w http.ResponseWriter, r *http.Request, _ httpr
 	h.d.Writer().WriteCreated(w, r, RouteBase+"?"+q.Encode(), &rel)
 }
 
-// swagger:route DELETE /relation-tuples write deleteRelationTuple
+// swagger:route DELETE /relation-tuples write deleteRelationTuples
```
zepatrik, line 169: This will result in breaking changes in the SDK, right @aeneasr? The behavior changed a bit, but nothing that should break existing integrations.

proto/ory/keto/acl/v1alpha1/write_service.proto
```diff
@@ -17,6 +17,8 @@ option php_namespace = "Ory\\Keto\\Acl\\V1alpha1";
 service WriteService {
   // Writes one or more relation tuples in a single transaction.
   rpc TransactRelationTuples(TransactRelationTuplesRequest) returns (TransactRelationTuplesResponse);
+  // Deletes relation tuples based on relation query
+  rpc DeleteRelationTuples(DeleteRelationTuplesRequest) returns (DeleteRelationTuplesResponse);
```
zepatrik, line 21: FYI, as this is a new rpc it will not break existing integrations.

zepatrik (approved):
Thank you, this looks very good now :+1: 
I am really sorry for the late review.

zepatrik:
I actually just realized that splitting the `delete` and `delete-all` into two subcommands makes more sense, as it now reflects the two behaviors and will not break existing behavior. Also, with this it should be harder to delete stuff by mistake. Thought it is easier if I just quickly do that instead of annoying you again :sweat_smile:
Thanks again for your work there, appreciate :+1:

PR #808, opened 2022-01-07
docs: clarify that CLI remotes should be addresses and not URIs

zepatrik:
## Related issue(s)

https://ory-community.slack.com/archives/C012RBZFMDG/p1641419049030200


jim-obrien-orig (commented):
LGTM (woohoo)

Issue #792, opened 2021-12-05
Automatic Subject/Object encoding


g13013:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/keto/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/keto/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

We are willing to use Keto in our stack as we think it's a perfect fit for our projects. We've been following the progress of Keto since the earliest versions. Before version 0.6 there was an AWS-like implementation I believe, it was convenient for us as our implementations are also inspired by AWS techs, for example, we are using URNs in our applications to reference objects between services, the change in recent Keto specs are fine except one thing, [Objects/Subjects naming limitations](https://www.ory.sh/keto/docs/concepts/subjects#advanced-example-using-application-information-within-keto-subjects). forcing every application to translate objects/subjects usually referenced by ARN's to UUIDs or so is far from being ideal. We believe there is a little change in Keto that could remove this limitation without forcing applications to obscure their implementation to overcome this limitation.

### Describe your ideal solution

In fact, this limitation could easily be addressed in Keto itself. Keto could do the encoding/decoding of the object/subject before hitting the database. The encoding could be done by using SHA256 or UUID-5  (with tuple's namespace).

- Write: Request -> Encode -> Backend
- Read: Request -> Encode -> Backend

Also, the object and subject type and format could be relaxed to include objects as well, which could be transformed to JSON and then encoded.

So taking the example from the Doc:

A request to keto like:

```json
{
    "action": "insert",
    "relation_tuple": {
        "namespace": "network",
        "object": "tcp/22",
        "relation": "access",
        "subject_id":  {
            "subnet": "192.168.0.0/24",
            "office_hours": true
         }
}
```

Assuming the namespace `network`'s uuid is `4fee2932-561d-11ec-8a6d-ff46757a867c` the request could be translated using uuid-5 to the following before write/read:

```json
{
    "action": "insert",
[...]

PR #809, opened 2022-01-09
feat: Add mapping table for encoding subject into UUIDs

hperl:
This PR contains first work towards supporting automatic subject/object encodings into UUIDs, as discussed in #792.

## Related issue(s)
#792

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Future work

In order to complete the work for #792, the following items still need to be adressed (in this or future PRs):

- [ ] RFC: Should the new encoding be enabled unconditionally, or opt-in as suggested in https://github.com/ory/keto/issues/792#issue-1071582579?
- [ ] Use the new mapping methods in the API when reading or writing relation tuples.

### DIFF
--- a/internal/persistence/sql/migrations/single_table_test.go
+++ b/internal/persistence/sql/migrations/single_table_test.go
@@ -5,7 +5,6 @@ import (
 	"strconv"
 	"testing"
 
-	"github.com/gofrs/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -95,15 +94,6 @@ func TestToSingleTableMigrator(t *testing.T) {
 				require.Len(t, rts, 2)
 				assert.Equal(t, sID, rts[0])
 				assert.Equal(t, sSet, rts[1])
-			})
-
-			t.Run("case=uuid mapping", func(t *testing.T) {
-				id := uuid.Must(uuid.NewV4())
-				rep1 := "foo"
-				require.NoError(t, r.UUIDMappingManager().AddUUIDMapping(ctx, id, rep1))
-				rep2, err := r.UUIDMappingManager().LookupUUID(ctx, id)
-				assert.NoError(t, err)
-				assert.Equal(t, rep1, rep2)
 			})
 
 			t.Run("case=paginates", func(t *testing.T) {
--- a/internal/persistence/sql/migrations/single_table_test.go
+++ b/internal/persistence/sql/migrations/single_table_test.go
@@ -226,7 +226,7 @@ func TestToSingleTableMigrator_HasLegacyTable(t *testing.T) {
 
 				ln, err := m.LegacyNamespaces(ctx)
 				require.NoError(t, err)
-				assert.Equal(t, nspaces, ln)
+				assert.ElementsMatch(t, nspaces, ln)
 			})
 		})
 	}
--- a/internal/persistence/sql/migrations/single_table_test.go
+++ b/internal/persistence/sql/migrations/single_table_test.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/gofrs/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -94,6 +95,15 @@ func TestToSingleTableMigrator(t *testing.T) {
 				require.Len(t, rts, 2)
 				assert.Equal(t, sID, rts[0])
 				assert.Equal(t, sSet, rts[1])
+			})
+
+			t.Run("case=uuid mapping", func(t *testing.T) {
+				id := uuid.Must(uuid.NewV4())
+				rep1 := "foo"
+				require.NoError(t, r.UUIDMappingManager().AddUUIDMapping(ctx, id, rep1))
+				rep2, err := r.UUIDMappingManager().LookupUUID(ctx, id)
+				assert.NoError(t, err)
+				assert.Equal(t, rep1, rep2)
 			})
 
 			t.Run("case=paginates", func(t *testing.T) {
--- a/internal/persistence/sql/migrations/templates/20220110200400_create-uuid-mapping-table.up.sql
+++ b/internal/persistence/sql/migrations/templates/20220110200400_create-uuid-mapping-table.up.sql
@@ -1,7 +1,7 @@
 CREATE TABLE keto_uuid_mappings
 (
-    id                       UUID        NOT NULL,
-    string_representation    VARCHAR(64) NOT NULL CHECK (string_representation <> ''),
+    id                       UUID         NOT NULL,
+    string_representation    VARCHAR(255) NOT NULL CHECK (string_representation <> ''),
 
     PRIMARY KEY (id),
 
--- a/internal/persistence/sql/migrations/templates/20220110200400_create-uuid-mapping-table.up.sql
+++ b/internal/persistence/sql/migrations/templates/20220110200400_create-uuid-mapping-table.up.sql
@@ -1,12 +1,7 @@
 CREATE TABLE keto_uuid_mappings
 (
-    id                       UUID         NOT NULL,
-    string_representation    VARCHAR(255) NOT NULL CHECK (string_representation <> ''),
+    id                       UUID NOT NULL,
+    string_representation    TEXT NOT NULL CHECK (string_representation <> ''),
 
-    PRIMARY KEY (id),
-
-    -- enforce uniqueness
-    CONSTRAINT chk_keto_uuid_map_uniq UNIQUE (id, string_representation)
+    PRIMARY KEY (id)
 );
-
-CREATE INDEX keto_uuid_mappings_subject_ids_idx ON keto_uuid_mappings (string_representation);
--- a/internal/persistence/sql/migrations/templates/20220110200400_create-uuid-mapping-table.up.sql
+++ b/internal/persistence/sql/migrations/templates/20220110200400_create-uuid-mapping-table.up.sql
@@ -1,0 +1,12 @@
+CREATE TABLE keto_uuid_mappings
+(
+    id                       UUID        NOT NULL,
+    string_representation    VARCHAR(64) NOT NULL CHECK (string_representation <> ''),
+
+    PRIMARY KEY (id),
+
+    -- enforce uniqueness
+    CONSTRAINT chk_keto_uuid_map_uniq UNIQUE (id, string_representation)
+);
+
+CREATE INDEX keto_uuid_mappings_subject_ids_idx ON keto_uuid_mappings (string_representation);
--- a/internal/persistence/sql/migrations/uuid_mapping.go
+++ b/internal/persistence/sql/migrations/uuid_mapping.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/gobuffalo/pop/v6"
 	"github.com/gofrs/uuid"
-
 	"github.com/ory/x/sqlcon"
 
 	"github.com/ory/keto/internal/persistence/sql"
@@ -96,7 +95,7 @@ func (m *toUUIDMappingMigrator) migrateObject(ctx context.Context, rt *sql.Relat
 }
 
 func (m *toUUIDMappingMigrator) addUUIDMapping(ctx context.Context, value string) (id string, err error) {
-	uid, err := m.d.Persister().MappedUUID(ctx, value)
+	uid, err := m.d.Persister().ToUUID(ctx, value)
 	return uid.String(), err
 }
 
--- a/internal/persistence/sql/migrations/uuid_mapping.go
+++ b/internal/persistence/sql/migrations/uuid_mapping.go
@@ -7,17 +7,19 @@ import (
 	"github.com/gofrs/uuid"
 
 	"github.com/ory/keto/internal/persistence/sql"
+	"github.com/ory/x/sqlcon"
 )
 
 type (
 	toUUIDMappingMigrator struct {
-		d dependencies
+		d       dependencies
+		perPage int
 	}
 )
 
 // NewToUUIDMappingMigrator creates a new UUID mapping migrator.
 func NewToUUIDMappingMigrator(d dependencies) *toUUIDMappingMigrator {
-	return &toUUIDMappingMigrator{d: d}
+	return &toUUIDMappingMigrator{d: d, perPage: 100}
 }
 
 // MigrateUUIDMappings migrates to UUID-mapped subject IDs for all relation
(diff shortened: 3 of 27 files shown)


hperl:
@zepatrik the Cockroach migration test failed with `std_err: Could not apply migrations: ERROR: relation "keto_uuid_mappings" already exists (SQLSTATE 42P07)`, although the table is only created once in the migration scripts. Did I misunderstand how you do migrations?

https://app.circleci.com/pipelines/github/ory/keto/1737/workflows/da82ec86-7471-4ff4-ab64-77d32d1a91d3/jobs/11054?invite=true#step-110-785

internal/persistence/sql/uuid_mapping.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	nsUUID, err := uuid.FromString(ns.UUIDv5Namespace)
+	if err != nil {
+		// Is this the right thing to do? I assue that if the namespace
+		// has no valid UUID, the encoding should still work.
+		nsUUID = uuid.Nil
+	}
```
zepatrik, line 37: This will be obsolete with UUIDv4.

zepatrik (changes requested):
Thanks, this is a good starting point but I think some requirements were not clearly defined in the issue. Now that I see the first implementation iteration, I clarified some parts :wink:

internal/namespace/definitions.go
```diff
@@ -7,9 +7,10 @@ import (
 
 type (
 	Namespace struct {
-		ID     int32           `json:"id" db:"-" toml:"id"`
-		Name   string          `json:"name" db:"-" toml:"name"`
-		Config json.RawMessage `json:"config,omitempty" db:"-" toml:"config,omitempty"`
+		ID              int32           `json:"id" db:"-" toml:"id"`
+		Name            string          `json:"name" db:"-" toml:"name"`
+		UUIDv5Namespace string          `json:"uuid5_namespace,omitempty" db:"-" toml:"uuid5_namespace,omitempty"`
```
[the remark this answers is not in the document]
hperl: Sounds good!

hperl:
Thanks for the review. I addressed most of the points (I think), but I am still not sure where the best place is to actually do the migration as well as add the business logic that inserts a UUID in the mapping table if a non-UUID subject ID should be inserted.

zepatrik (commented):
@hperl thanks for addressing the feedback :+1: and sorry for the late reply :see_no_evil: 

I have moved the mapping manager test to the right location, I think it is easier to show by commit than to explain :wink: 
This is now pretty much what I expected as the first draft.
For the migration, we would ideally implement that in SQL. If it gets too complicated there (especially for cross-DB compatibility), we can also do it in go. There is already one migration process, the [`SingleTableMigrator`](https://github.com/ory/keto/blob/fabf1a06d90139f6678fc7bc1703ee92a5fc9d69/internal/persistence/sql/migrations/single_table.go) that runs in go. You can give it a try in SQL if you want :wink: 

In all the handlers/... we would then change the subject and object fields to be one of `subject_uuid` or `subject_string`. In the latter case, we would apply the mapping. But I would do that in a separate follow up PR if you don't want to invest all that work.

Oh and btw, for that we will also need a string to UUID function, and therefore also an (unique) index on the string column. It would be great if you can add that in this PR.

hperl:
@zepatrik, thanks for moving stuff to the correct place. It feels a little strange to have the test function defined outside of a `_test` file, though.

As for the SQL migration, that would still go into [20220110200400_create-uuid-mapping-table.up.sql](https://github.com/ory/keto/blob/f519a5c16c62e7194fb753ef7a066f06e5f9033c/internal/persistence/sql/migrations/templates/20220110200400_create-uuid-mapping-table.up.sql), correct?

zepatrik:
> It feels a little strange to have the test function defined outside of a _test file, though.

We use this pattern for the persister tests only. Main reason being that we can avoid circular dependencies with that. Also, we don't usually spin up all databases for all tests, but only have this one persister test that uses ALL supported databases.

And yes, you can try to add the migration in that file :wink:

hperl:
Having thought about writing the migration in SQL, I think Go is the better choice. The migration would need to parse a UUID that conforms to the Go implementation (https://pkg.go.dev/github.com/gofrs/uuid?utm_source=gopls#UUID.UnmarshalText) as well as generate UUIDs (and at least MySQL generates v1 UUIDs only, which are not random, see https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_uuid).

With both generation and parsing being hard to do consistently across SQL dialects, I'd go with Go. @zepatrik, WDYT?

rdurell:
If mapping subject to UUID, why not increase the supported subject length as well?  We're looking at using somewhat complex subject strings which would benefit by being larger then 64 characters.

hperl:
@rdurell sure, I increased it to 1024. Do you have more information on typical subject ID sizes?

zepatrik:
> With both generation and parsing being hard to do consistently across SQL dialects, I'd go with Go.

Sounds reasonable :+1: 

> why not increase the supported subject length as well?

Definitely, maybe we can even make the strings unbound? I know that at least postgres and cockroach don't really care. If you happen to store big strings, you will face the performance implications of that. But we should not force you to do that.

internal/persistence/sql/migrations/single_table_test.go
```diff
@@ -226,7 +226,7 @@ func TestToSingleTableMigrator_HasLegacyTable(t *testing.T) {
 
 				ln, err := m.LegacyNamespaces(ctx)
 				require.NoError(t, err)
-				assert.Equal(t, nspaces, ln)
+				assert.ElementsMatch(t, nspaces, ln)
```
zepatrik, line 229: Is this migration affected by your changes? It is meant to be removed on a future release anyways, so we can also do it now. Just needs some documentation on how to migrate then (i.e. first use Keto v0.7.0 and continue from there).

internal/persistence/sql/migrations/uuid_mapping.go
```diff
@@ -0,0 +1,56 @@
...
+func (m *toUUIDMappingMigrator) MigrateUUIDMappings(ctx context.Context) error {
+	p, ok := m.d.Persister().(*sql.Persister)
+	if !ok {
+		panic("got unexpected persister")
+	}
+
+	return p.Transaction(ctx, func(ctx context.Context, c *pop.Connection) error {
+		var relationTuples []*sql.RelationTuple
+
+		if err := p.Connection(ctx).All(&relationTuples); err != nil {
```
zepatrik, line 34: Getting all will probably not work for bigger databases. Please add pagination to this. You can use keyset pagination https://www.cockroachlabs.com/docs/stable/pagination.html#keyset-pagination for that. Should be the fastest one.

internal/persistence/sql/migrations/uuid_mapping.go
```diff
@@ -0,0 +1,56 @@
...
+
+			id := uuid.Must(uuid.NewV4())
+			if err := p.AddUUIDMapping(ctx, id, rt.SubjectID.String); err != nil {
+				return err
+			}
+
+			rt.SubjectID.String = id.String()
+			if err := c.Update(rt); err != nil {
+				return err
+			}
```
zepatrik, line 52: This does not cover all cases. We have to migrate all object and subject IDs that are not UUIDs.
In case the tuple has a subject set, we will have to migrate the tuple object and the subject set object. In case it has a subject ID, we have to map object and subject ID (as you already did).

internal/persistence/sql/migrations/uuid_mapping_test.go
```diff
@@ -0,0 +1,85 @@
...
+				name: "with null subject",
+				rt: &sql.RelationTuple{
+					ID:                    uuid.Must(uuid.NewV4()),
+					SubjectID:             dbsql.NullString{String: "", Valid: false},
+					SubjectSetNamespaceID: dbsql.NullInt32{Int32: 0, Valid: true},
+					SubjectSetObject:      dbsql.NullString{String: "obj", Valid: true},
+					SubjectSetRelation:    dbsql.NullString{String: "rel", Valid: true},
+					CommitTime:            time.Now(),
+				},
+				expectMapping: false,
```
zepatrik, line 51:
```suggestion
				expectMapping: true,
```

zepatrik (changes requested):
Thank you, just some remarks on the migrator.

internal/persistence/sql/migrations/single_table_test.go
hperl: It was a flaky test, since the order of the two slices was not guaranteed.

zepatrik: Ah I see :+1:

internal/persistence/sql/migrations/uuid_mapping.go
hperl: I now added pagination with `(pop.Query).Paginate`, which seems to be the cross-db way of doing things.

hperl:
> Definitely, maybe we can even make the strings unbound? I know that at least postgres and cockroach don't really care. If you happen to store big strings, you will face the performance implications of that. But we should not force you to do that.

We can either use a `VARCHAR(x)` with a large x, or even `TEXT`, but in both cases MySQL will limit us, because indices are limited in size (and we put an index on both the UUID and the string representation). So even for `TEXT` the index would then need to take a prefix length, see https://dev.mysql.com/doc/refman/8.0/en/create-index.html.

For storing arbitrary large strings, I would not want to put an index on them or even lookup by the string representation then, but either:
* add another string column that holds a hash of the string representation, put an index on that and hash the strings before lookup
* Generate the UUIDs with UUIDv5 directly from the string representation and a constant or empty namespace (<-- preferred)

Maybe we can quickly go over the last open points of the design in-band, I think we could quickly resolve them.

internal/persistence/sql/migrations/uuid_mapping.go
hperl: Makes sense, I added migration of objects as well.

internal/persistence/sql/uuid_mapping.go
```diff
@@ -24,17 +23,32 @@ func (UUIDMapping) TableName() string {
...
+	default:
+		query = `
+			INSERT INTO keto_uuid_mappings (id, string_representation)
+			VALUES (?, ?)
+			ON CONFLICT (id) DO NOTHING`
+	}
+
+	return id, sqlcon.HandleError(
+		p.Connection(ctx).RawQuery(query, id, text).Exec(),
+	)
```
zepatrik, line 48: You should be able to check for this error also using `sqlcon.HandleError` and `errors.Is`: https://github.com/ory/x/blob/08ea77b2a961bd23c170ddb076b10686c71e93f1/sqlcon/error.go#L18-L24

internal/persistence/sql/uuid_mapping.go
```diff
@@ -24,17 +23,32 @@ func (UUIDMapping) TableName() string {
 	return "keto_uuid_mappings"
 }
 
-func (p *Persister) AddUUIDMapping(ctx context.Context, id uuid.UUID, representation string) error {
-	m := &UUIDMapping{
-		ID:                   id,
-		StringRepresentation: representation,
-	}
+func (p *Persister) ToUUID(ctx context.Context, text string) (uuid.UUID, error) {
+	id := uuid.NewV5(p.NetworkID(ctx), text)
```
zepatrik, line 27: One more thing I realized: subjects are equal by string comparison, but objects are namespace bound. This means that the objects in the tuples `groups:root#member@foo` and `files:root#access@foo` are not referring to the same object, but the subjects are the same in those tuples.
Therefore I think we have to add the namespace to the UUID step as well. Not sure how though, because if we just string-append the namespace and object, it can collide with a subject that is that string. Additionally namespace names can change, but the namespace numeric ID is constant.

What do you think of
```suggestion
	id := uuid.NewV5(uuid.NewV5(p.NetworkID(ctx), strconv.Itoa(namespace.ID)), text)`
```
for objects, and yours for subjects?

internal/uuidmapping/definitions.go
```diff
@@ -15,37 +15,34 @@ type (
...
 	}
 	Manager interface {
-		// MappedUUID returns the mapped UUID for the given string
-		// representation. If the string representation is not mapped, a new
-		// UUID will be created automatically.
-		MappedUUID(ctx context.Context, representation string) (uuid.UUID, error)
+		// ToUUID returns the mapped UUID for the given string representation.
+		// If the string representation is not mapped, a new UUID will be
+		// created automatically.
+		ToUUID(ctx context.Context, representation string) (uuid.UUID, error)
```
zepatrik, line 21: We then here probably need two functions, one for objects and one for subjects.

internal/persistence/sql/uuid_mapping.go
hperl: Ah, thanks!

internal/persistence/sql/uuid_mapping.go
hperl: Is that really an issue? I assumed that the objects would be considered namespace-bound elsewhere, so even if the same UUID appears in multiple namespaces (as now arbitrary strings would), this would not be a problem.

In summary, I would rather take the namespace into consideration for object equivalence rather than relying on them being mapped to different UUIDs.

zepatrik: True, good point :+1:

hperl: OK nice, I think we have everything resolved then :)

internal/persistence/sql/uuid_mapping.go
hperl: I just tried that, and one big difference between telling the db that it should ignore the conflict and ignoring the conflict in Go is that in the former case, the transaction is cancelled.

zepatrik: I see, and we might want to run this as part of some outer transaction. Then I guess we have to keep that.

internal/persistence/sql/migrations/uuid_mapping_test.go
```diff
@@ -0,0 +1,102 @@
...
+							assert.Equal(t, newRt.Object, newRt.SubjectSetObject.String)
+						}
+						assertHasMapping(t, conn, tc.rt.Object, newRt.Object)
+					} else {
+						// Nothing should have changed (ignoring commit time)
+						newRt.CommitTime = tc.rt.CommitTime
+						assert.Equal(t, tc.rt, newRt)
+					}
+				})
+			}
```
zepatrik, line 89: I would like to add a test case here that ensures the pagination works. There is a wrapper for the relation tuple manager where you can overwrite the default pagination: https://github.com/ory/keto/blob/1e15176233eac5db982721e95604cd8e9d47e1a2/internal/relationtuple/definitions.go#L644-L660
You can initialize and use it in the test similar to https://github.com/ory/keto/blob/1e15176233eac5db982721e95604cd8e9d47e1a2/internal/check/engine_test.go#L34-L42
https://github.com/ory/keto/blob/1e15176233eac5db982721e95604cd8e9d47e1a2/internal/check/engine_test.go#L441

internal/persistence/sql/migrations/templates/20220110200400_create-uuid-mapping-table.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE keto_uuid_mappings
+(
+    id                       UUID NOT NULL,
+    string_representation    TEXT NOT NULL CHECK (string_representation <> ''),
+
+    PRIMARY KEY (id)
+);
```
zepatrik, line 7: I just realized that we of course have to change the column types in the relation tuple table to UUID as well. I am afraid that we have to create a new table and insert into that instead of the old one. Or do you have other ideas?
The migrator we use can theoretically be extended to add go migrations btw, I think that would definitely make sense. Then we also don't need the command :thinking:
What do you think about that @aeneasr?

https://github.com/ory/x/blob/fdc336251a395d29c7a242b611dc3fe3740620f8/popx/migration_box.go

zepatrik (commented):
I am really sorry for the long review time. This looks very good now.
IMO we could merge it as is and open new PRs to finish the feature? This would block us however on creating releases...
Alternatively, we can create a feature branch that we then open more PRs against and in the end rebase master on that.

internal/persistence/sql/migrations/templates/20220110200400_create-uuid-mapping-table.up.sql
hperl: Creating a new table might not be necessary. Assuming that `object` and `subject_id` already hold UUIDs, we can, in Postgres, do:
```sql
ALTER TABLE keto_relation_tuples ALTER object TYPE UUID using object::uuid;
```
(and similar for `subject_id`). MySQL does not have a UUID type, so we will stick with `VARCHAR(64)`. The same statement works in Cockroachdb when using `SET enable_experimental_alter_column_type_general = true`.

internal/persistence/sql/migrations/uuid_mapping_test.go
hperl: Thanks, I added a test for pagination. I did not use the `ManagerWrapper`, because I was not retrieving the tuples via the manager anyways, but I used that code as inspiration nonetheless 😉.

hperl:
> I am really sorry for the long review time. This looks very good now. IMO we could merge it as is and open new PRs to finish the feature? This would block us however on creating releases... Alternatively, we can create a feature branch that we then open more PRs against and in the end rebase master on that.

I'm not a big fan of long-living feature branches, but I don't know the release process well enough. If we could prevent the migration from running, nothing user-visible changed. Is that an option?

zepatrik:
> I'm not a big fan of long-living feature branches, but I don't know the release process well enough.

Agree, but we can only release master as is easily, so this would block us. You are right that currently it does not affect the previous code, but it will once we start integrating the mapping into handlers, ...
Just keeping this off of the main branch will make it a bit easier.

zepatrik (approved):

Issue #463, opened 2021-03-01
Add tracing and metrics (prometheus/jaeger/...) capabilities


zepatrik:
**Is your feature request related to a problem? Please describe.**

Currently there is no tracing middleware loaded.

**Describe the solution you'd like**

Instantiation of the servers:
- [ ] https://github.com/ory/keto/blob/5f25bceea35179c67d24dd95f698dc57b789d87a/internal/driver/registry_default.go#L227-L239
- [ ] https://github.com/ory/keto/blob/5f25bceea35179c67d24dd95f698dc57b789d87a/internal/driver/registry_default.go#L250-L262
- [ ] https://github.com/ory/keto/blob/5f25bceea35179c67d24dd95f698dc57b789d87a/internal/driver/registry_default.go#L212-L213
- [ ] https://github.com/ory/keto/blob/5f25bceea35179c67d24dd95f698dc57b789d87a/internal/driver/registry_default.go#L197-L198

**Additional context**

Kratos' prometheus endpoint implementation: https://github.com/ory/kratos/tree/master/metrics/prometheus
Other tracing options in Kratos can be found here: https://github.com/ory/kratos/blob/412af90a9cc86cb3b2fd8059b6fa51d640b23c05/driver/config/config.go

aeneasr:
@zepatrik jaeger and prometheus are doing two separate things:

- Jaeger enables tracing - including tracing is very easy, you can check the source code (especially SQL init) from hydra/kratos and just copy paste that
- Prometheus enables metrics - including is also very easy as there are already available metricsx packages in x available (you can also refer to hydra for that!)

Enjoy :)

PR #832, opened 2022-02-08
feat: add new metrics server to keto

Demonsthere:
This PRs adds a new http only metrics server to keto, which exposes the default prometheus metrics.
Changes in charts will be done in a separate PR afterwards

## Related issue(s)
- Followup to https://github.com/ory/keto/issues/463

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/internal/driver/daemon.go
+++ b/internal/driver/daemon.go
@@ -100,29 +100,16 @@ func (r *RegistryDefault) ServeMetrics(ctx context.Context) func() error {
 		})
 
 		eg.Go(func() error {
-			if err := s.Serve(l); !errors.Is(err, http.ErrServerClosed) {
-				// unexpected error
-				return errors.WithStack(err)
-			}
-			return nil
+			retrun graceful.Graceful(s.ListenAndServe, s.Shutdown)
 		})
-		eg.Go(func() error {
-			err := s.Serve(l)
-			if err != nil && !strings.Contains(err.Error(), "use of closed network connection") {
-				// unexpected error
-				return errors.WithStack(err)
-			}
-			// trigger further shutdown
-			cancel()
-			return nil
-		})
-
 		eg.Go(func() error {
 			<-ctx.Done()
 			ctx, cancel := context.WithTimeout(context.Background(), graceful.DefaultReadTimeout)
 			defer cancel()
 			return s.Shutdown(ctx)
 		})
+		
+		return eg.Wait()
 		return nil
 	}
 }
--- a/internal/driver/daemon.go
+++ b/internal/driver/daemon.go
@@ -4,7 +4,12 @@ import (
 	"context"
 	"net"
 	"net/http"
+	"os"
+	"os/signal"
 	"strings"
+	"syscall"
+
+	"github.com/ory/x/logrusx"
 
 	"github.com/ory/keto/internal/check"
 	"github.com/ory/keto/internal/expand"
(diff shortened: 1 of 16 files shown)


internal/driver/daemon.go
```diff
@@ -84,32 +84,42 @@ func (r *RegistryDefault) ServeWrite(ctx context.Context) func() error {
 	}
 }
 
+func (r *RegistryDefault) ServeMetrics(ctx context.Context) func() error {
+	rt := r.MetricsRouter()
+
+	return func() error {
+		return multiplexPortNoGRPC(ctx, r.Config().MetricsListenOn(), rt)
+	}
```
zepatrik, line 92: In cases were you don't have gRPC, you don't have to multiplex at all.

```suggestion
	return func() error {
		graceful.WithDefaults(&http.Server{
			Handler: r.MetricsRouter(),
			Addr:    r.Config().MetricsListenOn(),
		})
	}
```

Will make the code a lot less complex.

internal/driver/registry_default.go
```diff
@@ -128,6 +132,21 @@ func (r *RegistryDefault) Tracer() *tracing.Tracer {
 	return r.tracer
 }
 
+func (r *RegistryDefault) MetricsHandler() *prometheus.Handler {
+	if r.metricsHandler == nil {
+
```
zepatrik, line 137:
```suggestion
```

internal/driver/registry_default.go
```diff
@@ -334,6 +353,18 @@ func (r *RegistryDefault) WriteRouter() http.Handler {
...
 
+func (r *RegistryDefault) MetricsRouter() http.Handler {
+	n := negroni.New(reqlog.NewMiddlewareFromLogger(r.Logger(), "keto").ExcludePaths(prometheus.MetricsPrometheusPath))
+	router := &x.MetricsRouter{Router: httprouter.New()}
+
+	r.PrometheusManager().RegisterRouter(router.Router)
+	r.MetricsHandler().SetRoutes(router.Router)
+	n.UseHandler(router)
+	n.Use(r.PrometheusManager())
+
```
zepatrik, line 364: Here we are missing the CORS middleware, add it the same way as above:
```suggestion

	var handler http.Handler = n
	options, enabled := r.Config().CORS("metrics")
	if enabled {
		handler = cors.New(options).Handler(handler)
	}

```

zepatrik: And you will have to add that interface to the CORS config provider.

internal/e2e/helpers.go
```diff
@@ -46,6 +46,8 @@ func newInitializedReg(t testing.TB, dsn *dbx.DsnT, cfgOverwrites map[string]int
 		config.KeyReadAPIPort:       ports[0],
 		config.KeyWriteAPIHost:      "127.0.0.1",
 		config.KeyWriteAPIPort:      ports[1],
+		config.KeyMetricsHost:       "127.0.01",
```
zepatrik, line 49:
```suggestion
		config.KeyMetricsHost:       "127.0.0.1",
```

internal/e2e/full_suit_test.go
```diff
@@ -77,6 +84,14 @@ func Test(t *testing.T) {
 				if tc, ok := cl.(transactClient); ok {
 					t.Run(fmt.Sprintf("transactClient=%T", cl), runTransactionCases(tc, addNamespace))
 				}
+				t.Run("case=metrics", func(t *testing.T) {
```
zepatrik, line 87:
```suggestion

				t.Run("case=metrics is served", func(t *testing.T) {
```

internal/driver/registry_default.go
```diff
@@ -334,6 +353,18 @@ func (r *RegistryDefault) WriteRouter() http.Handler {
...
 }
 
+func (r *RegistryDefault) MetricsRouter() http.Handler {
+	n := negroni.New(reqlog.NewMiddlewareFromLogger(r.Logger(), "keto").ExcludePaths(prometheus.MetricsPrometheusPath))
+	router := &x.MetricsRouter{Router: httprouter.New()}
+
+	r.PrometheusManager().RegisterRouter(router.Router)
+	r.MetricsHandler().SetRoutes(router.Router)
+	n.UseHandler(router)
+	n.Use(r.PrometheusManager())
```
zepatrik, line 363: The router types are there to ensure on a compile level that we don't accidentally register write routes on the read endpoint. For metrics you can just use

```suggestion
	router := httprouter.New()

	r.PrometheusManager().RegisterRouter(router)
	r.MetricsHandler().SetRoutes(router)
	n.UseHandler(router)
	n.Use(r.PrometheusManager())
```

internal/x/router.go
```diff
@@ -11,4 +11,7 @@ type (
 	WriteRouter struct {
 		*httprouter.Router
 	}
+	MetricsRouter struct {
+		*httprouter.Router
+	}
```
zepatrik, line 16: And then this is not needed.

```suggestion
```

internal/driver/daemon.go
```diff
@@ -84,6 +84,49 @@ func (r *RegistryDefault) ServeWrite(ctx context.Context) func() error {
...
+			cancel()
+			return nil
+		})
+
+		eg.Go(func() error {
+			<-ctx.Done()
+			ctx, cancel := context.WithTimeout(context.Background(), graceful.DefaultReadTimeout)
+			defer cancel()
+			return s.Shutdown(ctx)
+		})
```
zepatrik, line 125: Here we only have one server, so you don't have to coordinate shutdown in case it errors. In the other case we start two servers and the multiplex listener, so it gets more complex. Here it is enough to just do

```suggestion
		eg.Go(func() error {
			retrun graceful.Graceful(s.ListenAndServe, s.Shutdown)
		})
		eg.Go(func() error {
			<-ctx.Done()
			ctx, cancel := context.WithTimeout(context.Background(), graceful.DefaultReadTimeout)
			defer cancel()
			return s.Shutdown(ctx)
		})
		
		return eg.Wait()
```

zepatrik (changes requested):
One small thing left :wink:

internal/e2e/full_suit_test.go
```diff
@@ -77,6 +84,22 @@ func Test(t *testing.T) {
 				if tc, ok := cl.(transactClient); ok {
 					t.Run(fmt.Sprintf("transactClient=%T", cl), runTransactionCases(tc, addNamespace))
 				}
+
+				t.Run("case=metrics is served", func(t *testing.T) {
```
Demonsthere, line 88:
```suggestion
				t.Run("case=metrics are served", func(t *testing.T) {
```

PR #862, opened 2022-03-26
refactor: API paths

zepatrik:
This change refactors the API paths to be consistent with the rest of the Ory ecosystem. This step is required for the unified Ory SDK. Additionally, as we plan to add high level APIs, e.g. for RBAC. The check and expand API paths changed to allow adding those.

BREAKING CHANGES: `/check` is now `/relation-tupes/check`
BREAKING CHANGES: `/expand` is now `/relation-tuples/expand`
BREAKING CHANGES: `/relation-tuples` is now `/admin/relation-tuples` for write APIs
BREAKING CHANGES: gRPC package is now called `ory.keto.relation_tuples.v1alpha2`
BREAKING CHANGES: gRPC relation-tuple-delta action enum names are prefixed with `ACTION_`


contrib/docs-code-samples/expand-api-display-access/01-expand-beach/curl.sh
```diff
@@ -6,5 +6,5 @@ curl -G --silent \
      --data-urlencode "relation=access" \
      --data-urlencode "object=/photos/beach.jpg" \
      --data-urlencode "max-depth=3" \
-     http://127.0.0.1:4466/expand | \
+     http://127.0.0.1:4466/acl/expand | \
```
aeneasr, line 9: ACL for Access Control List? Is that the nomenclautur you want to go here?

zepatrik: I was also not sure about that, but basically yes. The gRPC API is also `package ory.keto.acl.v1alpha1`, so it would be consistent.
In some way you could say that relation tuples are an ACL?

zepatrik: Later on we can have `/rbac/check` and others.

Issue #263, opened 2020-10-05
[next-gen] Allow defining userset rewrites


zepatrik:
As defined in the paper §2.3.1

robinbraemer:
This is a [reference implementation](https://github.com/mishudark/zanzibar/blob/b11a1169349c61b43c5a09a1dcf6c1b31b038428/zanzibar.go#L94) for "tuple to subject set" rewrite rule.

Segflow:
Hello, we are interested in using Keto as our authorization system, and we will need this feature to models groups and access to groups. 

Can we get an update on this to evaluate if we should go forward with Keto or not?

zepatrik:
The update is that this is the highest priority feature as everyone needs it :wink:
But we cannot give out deliverable timelines for open source products. If you however want to support us and help us dedicate more resources four our open source work, reach out to hi@ory.sh

RichiCoder1:
![ItsHappeningGIF](https://user-images.githubusercontent.com/2391878/185688334-cd930678-8ed2-47fa-babe-fd8139a1bf7f.gif)

mojimi:
Hi, are there any examples on how to use this? And how to represent this in OPL?

PR #877, opened 2022-05-20
feat: user-set rewrites

hperl:
This PR implements userset rewrites as described in the Zanzibar paper (https://storage.googleapis.com/pub-tools-public-publication-data/pdf/10683a8987dbf0c6d4edcafb9b4f05cc9de5974a.pdf). This is work-in-progress.

The first part of the implementation focusses on checks against an internal representation of the namespace configuration, a configuration similar to the protobufs in the Zanzibar paper.

A later part would then translate from a high-level configuration language down to the internal representation.

## Related issue(s)

https://github.com/ory/keto/issues/263

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/internal/check/engine.go
+++ b/internal/check/engine.go
@@ -3,14 +3,16 @@ package check
 import (
 	"context"
 	"errors"
-
-	"github.com/ory/keto/internal/driver/config"
-	"github.com/ory/keto/internal/x/graph"
+	"time"
 
 	"github.com/ory/herodot"
 
+	"github.com/ory/keto/internal/driver/config"
+	"github.com/ory/keto/internal/namespace"
+	"github.com/ory/keto/internal/namespace/ast"
 	"github.com/ory/keto/internal/relationtuple"
 	"github.com/ory/keto/internal/x"
+	"github.com/ory/keto/internal/x/graph"
 )
 
 type (
@@ -25,6 +27,10 @@ type (
 		config.Provider
 		x.LoggerProvider
 	}
+
+	// Type aliases for shorter signatures
+	RelationTuple = relationtuple.InternalRelationTuple
+	Query         = relationtuple.RelationQuery
 )
 
 func NewEngine(d EngineDependencies) *Engine {
@@ -33,16 +39,18 @@ func NewEngine(d EngineDependencies) *Engine {
 	}
 }
 
-func (e *Engine) subjectIsAllowed(
+func (e *Engine) checkDirect(
 	ctx context.Context,
-	requested *relationtuple.InternalRelationTuple,
-	rels []*relationtuple.InternalRelationTuple,
+	requested *RelationTuple,
+	rels []*RelationTuple,
 	restDepth int,
-) (bool, error) {
+) checkFn {
 	// This is the same as the graph problem "can requested.Subject be reached from requested.Object through the first outgoing edge requested.Relation"
 	//
 	// We implement recursive depth-first search here.
 	// TODO replace by more performant algorithm: https://github.com/ory/keto/issues/483
+
+	var indirectChecks []checkFn
 
 	for _, sr := range rels {
 		ctx, wasAlreadyVisited := graph.CheckAndAddVisited(ctx, sr.Subject)
@@ -53,7 +61,7 @@ func (e *Engine) subjectIsAllowed(
 		// we only have to check Subject here as we know that sr was reached from requested.ObjectID, requested.Relation through 0...n indirections
 		if requested.Subject.Equals(sr.Subject) {
 			// found the requested relation
-			return true, nil
+			return isMemberCheckFn
 		}
 
 		sub, isSubjectSet := sr.Subject.(*relationtuple.SubjectSet)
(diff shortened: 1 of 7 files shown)


aeneasr:
Already? Wow!

aeneasr (commented):
Overall I find the code very enjoyable to read. I have only found some minor things / questions. I haven't looked into the algorithm itself yet.

@hperl if you decide to venture into caching, let's please synch first. At Ory, we've been badly burned by caching in security modules and there are several considerations we need to make to have a consistent and in particular secure cache! I'd also be interested to see a couple of benchmark tests to understand what we can improve where :)

internal/check/engine.go
```diff
@@ -33,91 +40,475 @@ func NewEngine(d EngineDependencies) *Engine {
...
-		prevPage = nextPage
+		resultCh <- g.Result()
 	}
 }
 
-func (e *Engine) SubjectIsAllowed(ctx context.Context, r *relationtuple.InternalRelationTuple, restDepth int) (bool, error) {
-	// global max-depth takes precedence when it is the lesser or if the request max-depth is less than or equal to 0
+func (e *Engine) SubjectIsAllowed(ctx context.Context, r *RelationTuple, restDepth int) (bool, error) {
+	result := e.Check(ctx, r, restDepth)
+	return result.Membership == checkgroup.IsMember, result.Err
```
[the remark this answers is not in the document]
hperl: Makes sense, will fix!

internal/check/engine.go
```diff
@@ -33,91 +40,475 @@ func NewEngine(d EngineDependencies) *Engine {
...
+		return checkgroup.Result{Err: context.Canceled}
+	}
+}
+
+func (e *Engine) checkIsAllowed(ctx context.Context, r *RelationTuple, restDepth int) checkgroup.Func {
+	if restDepth < 0 {
+		e.d.Logger().
+			Debug("reached max-depth, therefore this query will not be further expanded")
+		return func(_ context.Context, resultCh chan<- checkgroup.Result) {
+			resultCh <- checkgroup.Result{Membership: checkgroup.MembershipUnknown}
```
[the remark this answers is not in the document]
hperl: Each check function writes to the resultCh exactly once, and the caller must ensure that the result is read. If the context is cancelled, the check function must write `Result{Err: ctx.Err()}` to the channel.

Not receiving from the channel would be a resource leak. In https://github.com/ory/keto/blob/a400364880fe6f34c89c8b90424ad529d74c15dc/internal/check/checkgroup/concurrent_checkgroup.go#L52 we explicitly receive all running go routines.

Alternatively, we could change the semantics to say that the check function must not send on the `resultCh` if the `ctx` was cancelled. WDYT?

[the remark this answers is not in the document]
hperl: I since discovered go.uber.org/goleak, which checks that there are no more goroutines running at the end of a test. I'll add this, so that we can be sure that nothing leaks or hangs.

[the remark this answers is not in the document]
hperl: Done, assumptions are now thoroughly tested :).

internal/check/engine.go
```diff
@@ -33,91 +40,475 @@ func NewEngine(d EngineDependencies) *Engine {
...
+	ctx context.Context,
+	r *RelationTuple,
+	userset *ast.ComputedUserset,
+	restDepth int,
+) checkgroup.Func {
+	if restDepth < 0 {
+		e.d.Logger().
+			Debug("reached max-depth, therefore this query will not be further expanded")
+		return func(_ context.Context, resultCh chan<- checkgroup.Result) {
+			resultCh <- checkgroup.Result{Membership: checkgroup.MembershipUnknown}
```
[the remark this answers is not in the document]
hperl: Done, assumptions are now thoroughly tested :).

hperl:
@zepatrik let's merge that soon before I have to rebase again :)

zepatrik (approved):

internal/check/binop.go
```diff
@@ -0,0 +1,80 @@
...
+func or(ctx context.Context, checks []checkgroup.CheckFunc) checkgroup.Result {
+	if len(checks) == 0 {
+		return checkgroup.ResultNotMember
+	}
+
+	resultCh := make(chan checkgroup.Result, len(checks))
+	childCtx, cancelFn := context.WithCancel(ctx)
+	defer cancelFn()
+
+	for _, check := range checks {
```
aeneasr, line 24: Spawning a potential unlimited amount of go routines that execute very little logic or which cause a lot of load should be avoided at all cost.

Calling things asynchronous makes sense if the workload is worth the overhead. If we are spawning a go routine to return a (more or less static) value, the overhead of spawning a goroutine is not worth it.

When using concurrency to execute costly functions - usually functions that use RPC calls such as SQL, redis, protobuf, ... - we need to limit how much load a single node can generate.

Please re-evaluate the concurrency pattern here and decide whether it is needed or not. If it is needed, please replace this with a configurable worker pool. If you need inspiration: https://github.com/aeneasr/pool

internal/check/binop.go
```diff
@@ -0,0 +1,80 @@
...
+		go check(childCtx, resultCh)
```
aeneasr, line 53: See comment above

internal/check/binop.go
zepatrik: Good point. We have implemented it this way because at a later point we want to off-load some of the concurrent processing to other Keto nodes, essentially a cluster mode.

hperl: The parallelism of the binary operations is directly bound by the branching of the AST. So if we put an upper bound on the number of expressions e.g., 'or' together, we automatically limit this here.

I am talking about expressions like this:
```ts
this.related.P.traverse(p => p.permits.view(ctx)) ||
this.related.A.includes(ctx.subject) ||
this.related.B.includes(ctx.subject) ||
this.related.C.includes(ctx.subject) ||
this.related.D.includes(ctx.subject)
```
which will result in all 5 checks executing in parallel.

I still think that executing these in parallel is warranted, since each check will ultimately execute one or more database queries.

I think the better approach is to limit the branching in the AST (and leave this code as is). Right now, only the depth of the AST is limited.

WDYT?

aeneasr: While that might limit the concurrency of a singular check, it won't limit the concurrency of millions of checks running in a multi-tenant system. If enough requests hit, and parallelism is not bound by a worker pool, we will run into DoS issues.

I thus recommend implementing a worker pool regardless of the decisions in the AST. Limiting depth and branching is a good idea in general as an unbound model could also lead to DoS attacks against Ory Cloud (by using a very wide policy or potentially recursive policy)

hperl: OK makes sense. This will then be a global worker pool for all checks, right? Not scoped to one top-level check request?

hperl: @aeneasr I implemented a worker pool (basically https://gobyexample.com/worker-pools), PTAL :smiley:

zepatrik:
Rewrote the history a bit, see https://github.com/ory/keto/tree/feat/userset-rewrites-backup for the old history.

PR #878, opened 2022-05-25
feat(tracing): improved tracing for persisters and requests

icyphox:
This patch introduces tracing for persisters and requests.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


internal/persistence/sql/relationtuples.go
```diff
@@ -176,6 +191,9 @@ func (p *Persister) whereSubject(ctx context.Context, q *pop.Query, sub relation
 }
 
 func (p *Persister) whereQuery(ctx context.Context, q *pop.Query, rq *relationtuple.RelationQuery) error {
+	ctx, span := p.d.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.whereQuery")
```
hperl, line 194: Where all are we adding tracing? I'm genuinely curious, as I haven't worked much with tracing before. This function will not perform any DB or IO actions itself but just create the query. Do we also trace these intermediate functions, or is it more for DB interactions?

internal/x/registry.go
```diff
@@ -12,3 +15,7 @@ type LoggerProvider interface {
 type WriterProvider interface {
 	Writer() herodot.Writer
 }
+
+type TracingProvider interface {
```
hperl, line 19: Small nit, would `TracerProvider` fit better to the existing provider naming scheme?

internal/persistence/sql/relationtuples.go
```diff
@@ -41,6 +41,9 @@ func (RelationTuple) TableName(_ context.Context) string {
 }
 
 func (r *RelationTuple) toInternal(ctx context.Context, nm namespace.Manager, p *Persister) (*relationtuple.InternalRelationTuple, error) {
+	ctx, span := p.d.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.toInternal")
```
hperl, line 44: I see that we are repeating the package and function name in this and other functions. I wonder if it would make sense to ask the runtime for that information instead, so that this information does not get out-of-sync in case we rename or move the functions. WDYT?

```go
func spanNameOfCurrentFunction() string {
	pc, _, _, ok := runtime.Caller(0)
	if !ok {
		return "unknown"
	}
	if f := runtime.FuncForPC(pc); f != nil {
		return f.Name()
	}
	return "unknown"
}
```

hperl (approved):

hperl (approved):

internal/persistence/sql/relationtuples.go
icyphox: Yes, we only want to trace DB/IO or other "slow" operations to help pin-point potential performance bottlenecks. In this case — I must admit — I didn't pay close attention to what the function was actually doing and simply created a child span. While it does no harm existing, I suppose I'll go ahead and remove it for cleanliness' sake. :)

internal/x/registry.go
icyphox: Ideally yes, but we've stuck to `TracingProvider` across the board everywhere else (in Backoffice, Kratos, ...) so I suppose we can leave it as is for consistency.

internal/persistence/sql/relationtuples.go
icyphox: Nifty, but am I using it correctly? (Looks to be returning its own name): https://go.dev/play/p/Aa0yIFb3cjq
I remember experimenting with this idea previously, this was my PoC:

```go
func spanName() {
	pc := make([]uintptr, 15)
	n := runtime.Callers(2, pc)
	frames := runtime.CallersFrames(pc[:n])
	frame, _ := frames.Next()
	fmt.Println(frame.Function)
}

// Prints:
// module/package.Foo
```

Perhaps I'll create a separate PR for it (should be in ory/x anyway).

icyphox: Ah, it should be `runtime.Caller(1)` in your example.

internal/x/registry.go
hperl: Ah thanks, makes sense

internal/driver/registry_default.go
```diff
@@ -114,6 +114,10 @@ func (r *RegistryDefault) Tracer(ctx context.Context) *otelx.Tracer {
 		r.tracer = t
 	}
 
+	if r.tracer.Tracer() == nil {
```
zepatrik, line 117: How can we run into this case? :thinking: 
https://github.com/ory/x/blob/32ada8266273e76e35cc692a3999eb966cbeab8b/otelx/otel.go#L64-L65

internal/driver/daemon.go
```diff
@@ -129,7 +128,7 @@ func (r *RegistryDefault) serveRead(ctx context.Context, done chan<- struct{}) f
 	rt, s := r.ReadRouter(ctx), r.ReadGRPCServer(ctx)
 
 	if tracer := r.Tracer(ctx); tracer.IsLoaded() {
-		rt = otelx.NewHandler(rt, "internal.driver.serveRead")
+		rt = x.TraceHandler(rt)
```
zepatrik, line 131:
```suggestion
		rt = otelx.TraceHandler(rt)
```

internal/driver/daemon.go
```diff
@@ -141,7 +140,7 @@ func (r *RegistryDefault) serveWrite(ctx context.Context, done chan<- struct{})
 	rt, s := r.WriteRouter(ctx), r.WriteGRPCServer(ctx)
 
 	if tracer := r.Tracer(ctx); tracer.IsLoaded() {
-		rt = otelx.NewHandler(rt, "internal.driver.serveWrite")
+		rt = x.TraceHandler(rt)
```
zepatrik, line 143:
```suggestion
		rt = otelx.TraceHandler(rt)
```

zepatrik (changes requested):
Nice, just some minor questions.
Is there maybe a linter we can use to ensure the span name and function name are consistent?

internal/driver/registry_default.go
icyphox: Ah I forgot I'd added a check for it already. I'll remove it here.

PR #886, opened 2022-06-21
ci: add dependabot updates

zepatrik:
Enables dependabot updates.


hperl (approved):
Thanks for enabling this!

PR #920, opened 2022-06-27
Feat/userset-rewrites-config

hperl:
This PR adds a lexer and parser for the Ory Permission Language.


hperl:
@zepatrik design of the lexer and parser is already ready for review/feedback, if you have the bandwidth :).

PR #937, opened 2022-07-05
fix: make dockle optional

hperl:
Fix Dockle as in https://github.com/goodwithtech/dockle/issues/188.


zepatrik (approved):

zepatrik (changes requested):
That still did not fix it, I think you have to remove that step completely (and `make format`) :wink:

zepatrik (approved):

PR #939, opened 2022-07-06
fix: enable telemetry by default

zepatrik:
The problem was that `strconv.ParseBool(os.Getenv("SQA_OPT_OUT"))` would result in an error when the env var would not be set (default), and therefore the default behavior is to disable telemetry.
The env var is handled inside of the metricsx package anyways: https://github.com/ory/x/blob/f721c4ee23a6965ae39dc2e99c0ff2f4eb161410/metricsx/middleware.go#L185


aeneasr:
Nice find, thank you :)

PR #971, opened 2022-07-29
refactor: change pagination to use keyset pagination

### LINKED ISSUE #797: ListIdentities Pagination Starts at 1 instead of 0
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/keto/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/keto/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

- `ListIdentities` pagination starts at index 1 instead of index 0. Page 0 is the same as page 1.
- `ListIdentities` output gets truncated - there's some weird behavior when removing roles for one user, a role would suddenly pop up for another user, which makes me believe that there’s some maximum constraint. I don’t think we’re really returning that much data though, only around 30 users and 3-4 roles per user. This behavior became normal after we removed a lot of role assignments

### Reproducing the bug

https://user-images.githubusercontent.com/85463385/146244909-c4cac2e3-d417-4760-bfe9-5bfbc0ca1037.mov

Conversation with Patrik:
- https://ory-community.slack.com/archives/C012RBZFMDG/p1639429238101900
- https://ory-community.slack.com/archives/C012RBZFMDG/p1639455679103600

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

0.6.7

### On which operating system are you observing this issue?

macOS

### In which environment are you deploying?

Kubernetes with Helm

### Additional Context

_No response_

zepatrik:
The page token now is the last ID of the previous page.  This enables faster queries and more stable pagination.
NOTE: in case an integration modified page tokens to control pagination, this change will break the integration. Page tokens are opaque strings and should never be messed with.

Closes #797


internal/persistence/sql/relationtuples.go
```diff
@@ -210,8 +210,9 @@ func (p *Persister) GetRelationTuples(ctx context.Context, query *relationtuple.
 	}
 
 	sqlQuery := p.QueryWithNetwork(ctx).
-		Order("nid, namespace, object, relation, subject_id, subject_set_namespace, subject_set_object, subject_set_relation, commit_time").
-		Paginate(pagination.Page, pagination.PerPage)
+		Order("shard_id, nid").
```
hperl, line 213: Since `shard_id` is already unique, just ordering by that should suffice, right?

hperl, line 213:
```suggestion
		Order("shard_id").
```

zepatrik: Generally yes, but the primary key is `("shard_id, "nid")`. Not all databases use index prefixes, so using this ensures good performance across all of them.

hperl (approved):

internal/persistence/sql/relationtuples.go
hperl: Thanks!

PR #984, opened 2022-08-10
feat: namespace config

hperl:
The userset rewrites can now be configured through the Ory Permission
Language (OPL), which is a subset of TypeScript. The OPL config is
referenced in the central configuration under namespaces as such:

  [...]
  namespaces:
    config: <location>
  [...]

The <location> can be any valid file, directory or URI.

This is a follow up to expose the features implemented in #877

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist/
 **/*.sqlite-journal
 .vscode/
 .fuzzer/
+keto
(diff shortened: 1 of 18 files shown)


zepatrik (approved):
@hperl note that I changed the `namespaces.config` key to `namespaces.location`

Issue #850, opened 2022-03-07
Check CLI should allow passing subject sets


zepatrik:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/keto/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/keto/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

As described in #815 it is currently not possible to pass subject sets as the subject in the check CLI command.

### Describe your ideal solution

Ideally we would have a flag to distinguish subject IDs and subject sets.

### Workarounds or alternatives

Alternative: Don't support it.
Workaround: Use expand.

### Version

all

PR #985, opened 2022-08-14
fix(cli): distinguish between subjectSet and subjectID in check command

treilik:
attempt of fixing #850 but the command:

`./keto check 'videos:/cats/1.mp4#owner' view videos '/cats/1.mp4'`
still Outputs "Denied" even though it should be "Allowed" because of:
```
NAMESPACE       OBJECT          RELATION NAME   SUBJECT
videos          /cats/1.mp4     view            videos:/cats/1.mp4#owner

```

### DIFF
--- a/cmd/check/root.go
+++ b/cmd/check/root.go
@@ -24,7 +24,10 @@ func (o *checkOutput) String() string {
 	return "Denied\n"
 }
 
-const FlagMaxDepth = "max-depth"
+const (
+	FlagSubjectSet = "subject-set"
+	FlagMaxDepth   = "max-depth"
+)
 
 func newCheckCmd() *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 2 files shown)


cmd/check/root.go
```diff
@@ -42,13 +44,35 @@ func newCheckCmd() *cobra.Command {
...
 
+			subject := args[0]
+			relation := args[1]
+			namespace := args[2]
+			object := args[3]
+
+			var s *rts.Subject
+
+			// distinguish between subjectSet and subjectID
+			if strings.Contains(subject, "#") {
```
hperl, line 55: I'd rather distinguish based on `:` here, since the relation could be empty (= wildcard) for a subject set, e.g., `check users:Foo access files README.md`.

hperl:
Hi, and thanks for your contribution! It looks quite solid already, but could you add some tests to see that it actually works? Thanks!

treilik:
> Hi, and thanks for your contribution! It looks quite solid already, but could you add some tests to see that it actually works? Thanks!

Hi, Yes I'll do :)

ketoapi/enc_proto.go
```diff
@@ -57,6 +57,15 @@ func (r *RelationTuple) ToProto() *rts.RelationTuple {
 	return res
 }
 
+func (s *SubjectSet) ToProto() *rts.SubjectSet {
```
zepatrik, line 60: It seems this is not used? I'd recommend to remove it again then :wink:

internal/e2e/cases_test.go
```diff
@@ -28,6 +28,38 @@ func runCases(c client, m *namespaceTestManager) func(*testing.T) {
...
+				Relation:  rel,
+			}
+			rt := &ketoapi.RelationTuple{
+				Object:     obj,
+				SubjectSet: ss,
+				Namespace:  n.Name,
+				Relation:   rel,
+			}
+
+			assert.True(t, c.check(t, rt))
```
zepatrik, line 61: The reason why this test is failing is that the created tuples are not related to the checked tuple. It would be sufficient here to just create one tuple with a subject set and then check for the same tuple.

cmd/check/root.go
```diff
@@ -42,13 +44,35 @@ func newCheckCmd() *cobra.Command {
...
+				su := &ketoapi.SubjectSet{}
+				su, err := su.FromString(subject)
+				if err != nil {
+					return err
+				}
+
+				s = rts.NewSubjectSet(su.Namespace, su.Object, su.Relation)
+			} else {
+				s = rts.NewSubjectID(subject)
+			}
```
zepatrik, line 65: I'd prefer to make this a small helper function that has its own unit test. That way we can specifically test for the subject parsing of the CLI with more edge cases. The e2e test makes sense as it is right now as well just to ensure that subject sets are parsed and supported by all clients.

zepatrik (changes requested):
Thanks, sorry for the late review. Had a few out-of-office days.
There is just a few small things needed for the test to pass :wink:

PR #988, opened 2022-08-18
fix: use TLS in gRPC client

hperl:
Enable TLS and certificate checking in the gRPC client when communicating with remote hosts.

### DIFF
--- a/cmd/client/grpc_client.go
+++ b/cmd/client/grpc_client.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"net"
 	"os"
 	"strings"
 	"time"
@@ -12,6 +13,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/client/grpc_client.go
```diff
@@ -62,7 +64,20 @@ func Conn(ctx context.Context, remote string) (*grpc.ClientConn, error) {
...
-	return grpc.DialContext(ctx, remote, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock(), grpc.WithDisableHealthCheck())
+	return grpc.DialContext(ctx, remote,
+		grpc.WithTransportCredentials(transportCredentials(remote)),
+		grpc.WithBlock(),
+		grpc.WithDisableHealthCheck())
+}
+
+func transportCredentials(remote string) credentials.TransportCredentials {
+	u, err := url.Parse(remote)
+	if err == nil && (u.Hostname() == "127.0.0.1" || u.Hostname() == "localhost") {
```
zepatrik, line 75: I think you rather have to use `net.SplitHostPort`, as `127.0.0.1` is likely not parsed correctly...
https://go.dev/play/p/80PCAF8EtBL

zepatrik (changes requested):
Should be an easy fix :wink:

zepatrik (approved):
Looks good :+1:

PR #991, opened 2022-08-19
feat: add bearer token auth

hperl:
This PR adds bearer token authorization to Keto.

### DIFF
--- a/cmd/check/root_test.go
+++ b/cmd/check/root_test.go
@@ -15,7 +15,7 @@ func TestCheckCommand(t *testing.T) {
 	defer ts.Shutdown(t)
 
 	stdOut := ts.Cmd.ExecNoErr(t, "subject", "access", nspace.Name, "object",
-		"--insecure-skip-hostname-verification=true",
+		"--insecure-skip-hostname-verification",
 	)
 	assert.Equal(t, "Denied\n", stdOut)
 }
--- a/cmd/check/root_test.go
+++ b/cmd/check/root_test.go
@@ -14,6 +14,8 @@ func TestCheckCommand(t *testing.T) {
 	ts := client.NewTestServer(t, client.ReadServer, []*namespace.Namespace{nspace}, newCheckCmd)
 	defer ts.Shutdown(t)
 
-	stdOut := ts.Cmd.ExecNoErr(t, "subject", "access", nspace.Name, "object")
+	stdOut := ts.Cmd.ExecNoErr(t, "subject", "access", nspace.Name, "object",
+		"--insecure-skip-hostname-verification=true",
+	)
 	assert.Equal(t, "Denied\n", stdOut)
 }
--- a/cmd/client/grpc_client.go
+++ b/cmd/client/grpc_client.go
@@ -30,7 +30,7 @@ const (
 
 	EnvReadRemote  = "KETO_READ_REMOTE"
 	EnvWriteRemote = "KETO_WRITE_REMOTE"
-	EnvAuthToken   = "ORY_PAT"
+	EnvAuthToken   = "KETO_BEARER_TOKEN"
 
 	ContextKeyTimeout contextKeys = "timeout"
 )
--- a/cmd/client/grpc_client.go
+++ b/cmd/client/grpc_client.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"net"
 	"os"
@@ -9,22 +10,26 @@ import (
 	"time"
 
 	"github.com/ory/x/flagx"
+	"golang.org/x/oauth2"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/credentials/oauth"
 )
 
 type contextKeys string
 
 const (
-	FlagReadRemote  = "read-remote"
-	FlagWriteRemote = "write-remote"
+	FlagReadRemote        = "read-remote"
+	FlagWriteRemote       = "write-remote"
+	FlagInsecureTransport = "insecure"
 
 	EnvReadRemote  = "KETO_READ_REMOTE"
 	EnvWriteRemote = "KETO_WRITE_REMOTE"
+	EnvAuthToken   = "ORY_PAT"
 
 	ContextKeyTimeout contextKeys = "timeout"
 )
@@ -48,15 +53,27 @@ func getRemote(cmd *cobra.Command, flagRemote, envRemote string) (remote string)
 	return remote
 }
 
+func getToken(cmd *cobra.Command) (token string) {
+	return os.Getenv(EnvAuthToken)
+}
+
 func GetReadConn(cmd *cobra.Command) (*grpc.ClientConn, error) {
-	return Conn(cmd.Context(), getRemote(cmd, FlagReadRemote, EnvReadRemote))
+	return Conn(cmd.Context(),
+		getRemote(cmd, FlagReadRemote, EnvReadRemote),
+		getToken(cmd),
+		flagx.MustGetBool(cmd, FlagInsecureTransport),
+	)
 }
 
 func GetWriteConn(cmd *cobra.Command) (*grpc.ClientConn, error) {
-	return Conn(cmd.Context(), getRemote(cmd, FlagWriteRemote, EnvWriteRemote))
+	return Conn(cmd.Context(),
+		getRemote(cmd, FlagWriteRemote, EnvWriteRemote),
+		getToken(cmd),
+		flagx.MustGetBool(cmd, FlagInsecureTransport),
+	)
 }
 
-func Conn(ctx context.Context, remote string) (*grpc.ClientConn, error) {
+func Conn(ctx context.Context, remote, token string, insecureTransport bool) (*grpc.ClientConn, error) {
 	timeout := 3 * time.Second
 	if d, ok := ctx.Value(ContextKeyTimeout).(time.Duration); ok {
 		timeout = d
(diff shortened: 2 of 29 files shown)


cmd/status/root_test.go
```diff
@@ -87,3 +91,40 @@ func TestStatusCmd(t *testing.T) {
 		})
 	}
 }
+
+func authInterceptor(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp any, err error) {
```
zepatrik, line 95: As this is just one way to do auth, I bet we will get tons of requests to add more once we add this one :sweat_smile: 
Let's discuss this offline next week.

cmd/client/grpc_client.go
```diff
@@ -48,39 +51,86 @@ func getRemote(cmd *cobra.Command, flagRemote, envRemote string) (remote string)
 	return remote
 }
 
+func getToken(cmd *cobra.Command) (token string) {
+	t, err := cmd.Flags().GetString(FlagAuthToken)
```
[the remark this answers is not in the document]
hperl: Good catch! I think I'd just move it to an environment variable then.

cmd/status/root_test.go
zepatrik: Never mind, this is just test code :sweat_smile:

cmd/check/root_test.go
```diff
@@ -14,6 +14,8 @@ func TestCheckCommand(t *testing.T) {
 	ts := client.NewTestServer(t, client.ReadServer, []*namespace.Namespace{nspace}, newCheckCmd)
 	defer ts.Shutdown(t)
 
-	stdOut := ts.Cmd.ExecNoErr(t, "subject", "access", nspace.Name, "object")
+	stdOut := ts.Cmd.ExecNoErr(t, "subject", "access", nspace.Name, "object",
+		"--insecure-skip-hostname-verification=true",
```
zepatrik, line 18: I think this should be enough
```suggestion
		"--insecure-skip-hostname-verification",
```

cmd/status/root_test.go
```diff
@@ -26,6 +30,7 @@ func TestStatusCmd(t *testing.T) {
 			ts.Cmd.PersistentArgs = append(ts.Cmd.PersistentArgs, "--"+cmdx.FlagQuiet, "--"+FlagEndpoint, string(serverType))
 
 			t.Run("case=timeout,noblock", func(t *testing.T) {
+				t.Skip()
```
zepatrik, line 33: Revert?

cmd/status/root_test.go
```diff
@@ -34,12 +39,13 @@ func TestStatusCmd(t *testing.T) {
 			})
 
 			t.Run("case=noblock", func(t *testing.T) {
+				t.Skip()
```
zepatrik, line 42: Revert?

contrib/docs-code-samples/expand-api-display-access/00-create-tuples/cli.sh
```diff
@@ -18,6 +18,6 @@ files:/photos/beach.jpg#access@(directories:/photos#access)
 files:/photos/mountains.jpg#access@(files:/photos/mountains.jpg#owner)
 files:/photos/mountains.jpg#access@(directories:/photos#access)' | \
   keto relation-tuple parse - --format json | \
-    keto relation-tuple create - >/dev/null \
+    keto relation-tuple create --insecure-disable-transport-security - >/dev/null \
```
zepatrik, line 21: As these are part of the docs, is there any way we could use TLS, probably through self-signed certs? I assume it will be hard, but just to ensure it is :wink:

internal/driver/daemon.go
```diff
@@ -384,11 +384,19 @@ func (r *RegistryDefault) streamInterceptors(ctx context.Context) []grpc.StreamS
...
-	s := grpc.NewServer(
+func (r *RegistryDefault) newGrpcServer(ctx context.Context) *grpc.Server {
+	opts := []grpc.ServerOption{
 		grpc.ChainStreamInterceptor(r.streamInterceptors(ctx)...),
 		grpc.ChainUnaryInterceptor(r.unaryInterceptors(ctx)...),
-	)
+	}
+	if r.grpcTransportCredentials != nil {
+		opts = append(opts, grpc.Creds(r.grpcTransportCredentials))
+	}
```
zepatrik, line 394: We should rather enable TLS on the multiplexer (if that is even possible?)

internal/driver/registry_factory.go
```diff
@@ -62,23 +71,63 @@ func NewDefaultRegistry(ctx context.Context, flags *pflag.FlagSet, withoutNetwor
...
+	once sync.Once
+	cert *tls.Certificate
+	err  error
+}
+
+var sharedTestCert selfSignedCert
+
+func (s *selfSignedCert) generate() {
+	s.once.Do(func() {
+		key, err := ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
```
zepatrik, line 105: Would it be faster to use instead?
```suggestion
		key, err := ecdsa.GenerateKey(elliptic.P224(), rand.Reader)
```

internal/e2e/full_suit_test.go
```diff
@@ -72,8 +72,13 @@ func Test(t *testing.T) {
 					New: func() *cobra.Command {
 						return cmd.NewRootCmd(nil)
 					},
-					Ctx:            ctx,
-					PersistentArgs: []string{"--" + cliclient.FlagReadRemote, reg.Config(ctx).ReadAPIListenOn(), "--" + cliclient.FlagWriteRemote, reg.Config(ctx).WriteAPIListenOn(), "--" + cmdx.FlagFormat, string(cmdx.FormatJSON)},
+					Ctx: ctx,
+					PersistentArgs: []string{
+						"--" + cliclient.FlagReadRemote, reg.Config(ctx).ReadAPIListenOn(),
+						"--" + cliclient.FlagWriteRemote, reg.Config(ctx).WriteAPIListenOn(),
+						"--insecure-disable-transport-security=true",
```
zepatrik, line 79: Maybe we can run this e2e test using the self-signed cert?

zepatrik (changes requested):

cmd/client/grpc_client.go
```diff
@@ -23,12 +25,38 @@ const (
 	FlagReadRemote  = "read-remote"
 	FlagWriteRemote = "write-remote"
 
+	FlagInsecureNoTransportSecurity  = "insecure-disable-transport-security"
+	FlagInsecureSkipHostVerification = "insecure-skip-hostname-verification"
+
 	EnvReadRemote  = "KETO_READ_REMOTE"
 	EnvWriteRemote = "KETO_WRITE_REMOTE"
+	EnvAuthToken   = "ORY_PAT"
```
zepatrik, line 33: I would use a more generic name, as we might also rename the tokens in cloud and this is Ory cloud specific. What about `KETO_BEARER_TOKEN`?

cmd/check/root_test.go
hperl: Yea I thought so too, but something broke in the flag parsing then, it tried to swallow the next param.

contrib/docs-code-samples/expand-api-display-access/00-create-tuples/cli.sh
hperl: We could transition to TLS-by-default by using the self-signed approach, and then we would "just" skip hostname verification.

Is this what you had in mind?

zepatrik: I would just prefer to not have those insecure flags in the docs samples if possible :sweat_smile: So if there is some way to make that happen it would be highly appreciated, but I guess it will not be as easy.

internal/driver/registry_factory.go
hperl: Some quick benchmarking suggests we should go with P256.

```
goos: linux
goarch: amd64
pkg: github.com/ory/keto/internal/driver
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
BenchmarkCertificateGeneration/P256-8         	    6140	    188965 ns/op	   37140 B/op	     866 allocs/op
BenchmarkCertificateGeneration/P224-8         	    2575	    457371 ns/op	   39339 B/op	     927 allocs/op
BenchmarkCertificateGeneration/P384-8         	     705	   1803005 ns/op	   43231 B/op	     931 allocs/op
BenchmarkCertificateGeneration/P521-8         	     212	   5367458 ns/op	   47601 B/op	     936 allocs/op
```

cmd/client/grpc_client.go
hperl: :+1: I took `ORY_PAT` specifically because it is used in Cloud.

internal/driver/daemon.go
hperl: Yes, it is possible if we mux on the established HTTP connection, as seen here:

https://github.com/google/keytransparency/pull/205/commits/ad24e6d02c6b2cd5318183667cf995b56da0a42d#diff-fca3329b29f1d82d36a43d5e6f37764d88bf80c4ad17f769c7a3c135860d0981R121-R125

hperl: WDYT, should we push this into a follow-up issue? TLS support in Keto would be a new feature.

internal/e2e/full_suit_test.go
hperl: Then we'd have to solve the multiplexing issue (which I think is a new feature).

contrib/docs-code-samples/expand-api-display-access/00-create-tuples/cli.sh
hperl: As discussed, we'll update the docs.

internal/e2e/full_suit_test.go
hperl: Then we'd have to solve the multiplexing issue (which I think is a new feature).

contrib/docs-code-samples/expand-api-display-access/00-create-tuples/cli.sh
hperl: As discussed, we'll update the docs.

internal/driver/daemon.go
zepatrik: Yes separate PR sounds good :+1: 
Should we merge this one after TLS or before?

internal/e2e/full_suit_test.go
hperl: Then we'd have to solve the multiplexing issue (which I think is a new feature).

internal/driver/daemon.go
hperl: I'd merge now and do TLS as a separate feature.

zepatrik (approved):

aeneasr:
@hperl can you resolve the merge conflicts? then we'll merge :)

PR #1021, opened 2022-09-14
chore: bump deps

hperl:
This PR bumps various dependencies in Go modules.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/ory/graceful v0.1.3
 	github.com/ory/herodot v0.9.13
 	github.com/ory/jsonschema/v3 v3.0.7
-	github.com/ory/keto/proto v0.0.0-00010101000000-000000000000
+	github.com/ory/keto/proto v0.9.0-alpha.0.pre.0
```
zepatrik, line 25: We should keep `v0.0.0-00010101000000-000000000000` here. This ensures that the version is always properly replaced, because without a replace you get an error. Otherwise we have to always keep this up-to-date on releases, but the version is always tagged **after** the release. This makes it impossible to keep this updated.

```suggestion
	github.com/ory/keto/proto v0.0.0-00010101000000-000000000000
```

zepatrik (changes requested):

PR #1029, opened 2022-09-21
feat: OPL typescript library

zepatrik:
**IMPORTANT** do not squash but rebase to preserve deprecation note.


contrib/namespace-type-lib/index.d.ts
```diff
@@ -0,0 +1,34 @@
+/// <reference no-default-lib="true"/>
+
+declare interface Boolean {}
```
hperl, line 3: Where is `Boolean` used (compared to lowercase `boolean`)?

contrib/namespace-type-lib/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@ory/keto-namespace-types",
+  "version": "0.9.0-alpha.0",
```
hperl, line 3: Do we have to manage that version manually?

contrib/namespace-type-lib/test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+class Group implements Namespace {
+  related: {
+    members: (User | Group)[]
+  }
+
+  permits = {
+    isMember: (ctx: Context): boolean =>
+      this.related.members.traverse((m) =>
+        m instanceof User ? m == ctx.subject : m.permits.isMember(ctx),
```
hperl, line 18: Checking by `instanceof` is not part of the OPL. I agree that there is some usefulness, but shouldn't we implement it before we put it into tests?

hperl (approved):

contrib/namespace-type-lib/index.d.ts
zepatrik: If I understand TS correctly, it is only about the functions defined on the primitives (like `"foo".includes("bar")`)

contrib/namespace-type-lib/package.json
zepatrik: No, the CI release does that :wink: https://github.com/ory/keto/pull/1029/files#diff-690379e7730ef01d6a0d83d5590312f76fbb839b4642554752d757b98e85365aR27

contrib/namespace-type-lib/test.ts
zepatrik: The test is only to ensure that the typing works. With this `instanceof` conditional I could verify that the members are actually either of type `User` or `Group`. It really only is about the type library correctness.

Issue #1040, opened 2022-09-29
Ory Permission Language files need not to have imports to `@ory/keto-namespace-types`


zepatrik:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/keto/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/keto/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [x] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

An OPL config must not contain
```ts
import { ... } from '@ory/keto-namespace-types'
```
while it typically does.

### Reproducing the bug

Try to load a config containing the import.

### Version

v0.10.0

https://ory-community.slack.com/archives/C012RBZFMDG/p1664442459359059?thread_ts=1664306167.280199&cid=C012RBZFMDG

PR #1041, opened 2022-09-29
fix: properly lex imports in OPL

hperl:
## Related issue(s)

Closes https://github.com/ory/keto/issues/1040

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


internal/schema/lexer.go
```diff
@@ -307,7 +307,7 @@ func lexBlockComment(l *lexer) stateFn {
 func lexStringLiteral(l *lexer) stateFn {
 	r := l.next()
 	l.ignore()
-	l.acceptRun(digits + letters)
+	l.acceptRun(digits + letters + "@/-")
```
zepatrik, line 310: hm shouldn't we accept basically anything except unescaped `"` here anyways?

zepatrik (commented):
So are `import`s now considered string literals, but their value is not used? Don't really understand how this works now :sweat_smile:

internal/schema/lexer.go
hperl: Yes, we can do that.

hperl:
> So are `import`s now considered string literals, but their value is not used? Don't really understand how this works now sweat_smile

In the lexing phase, we just want to generate a token stream from the input. So `import { Foo, Bar, Baz } from '@ory/keto-namespace-types'` becomes this:

```json
[
  "'import'",
  "{",
  "'Foo'",
  ",",
  "'Bar'",
  ",",
  "'Baz'",
  "}",
  "'from'",
  "'@ory/keto-...'",
  "EOF"
]
```

In the parsing phase, this stream will be complete ignored because we just skip until we find a `class` keyword.

zepatrik (approved):

PR #1066, opened 2022-10-11
feat: expose OPL syntax check API

hperl:
We how have a new OPL API with a function to syntax-check an OPL file.

## Related issue(s)

Part of https://github.com/ory-corp/cloud/issues/2847

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.
- [x] I have added or changed [the documentation](https://github.com/ory/docs).


go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/ory/graceful v0.1.3
 	github.com/ory/herodot v0.9.13
 	github.com/ory/jsonschema/v3 v3.0.7
-	github.com/ory/keto/proto v0.0.0-00010101000000-000000000000
+	github.com/ory/keto/proto v0.10.0-alpha.0
```
zepatrik, line 25: Please revert, so there cannot be a mismatch.

internal/driver/config/provider.go
```diff
@@ -155,23 +156,39 @@ func (k *Config) Set(key string, v any) error {
...
 }
 
-func (k *Config) WriteAPIListenOn() string {
+func (k *Config) MetricsListenOn() string {
+	return fmt.Sprintf(
+		"%s:%d",
+		k.p.String(KeyMetricsHost),
+		k.p.Int(KeyMetricsPort),
+	)
+}
```
zepatrik, line 178: Maybe instead we could do

```suggestion
func (k *Config) listenOn(hostKey, portKey string) string {
	return fmt.Sprintf(
		"%s:%d",
		k.p.String(hostKey),
		k.p.Int(portKey),
	)
}

func (k *Config) MetricsListenOn() string {
	return k.listenOn(KeyMetricsHost, KeyMetricsPort)
}
```

zepatrik (changes requested):
Nice :+1: just some minor naming suggestions, and we should be ready to :bus:

internal/driver/config/provider.go
hperl: Yes, I'll refactor. That part was nagging me as well :)

zepatrik (approved):

PR #1145, opened 2022-11-22
fix: relative file URL parsing

zepatrik:
See https://github.com/ory/x/blob/a25499c90cdfa8d8b52298bf9e2d87e1078ec145/urlx/parse.go#L19-L23


hperl (approved):
:tada:

PR #1168, opened 2022-12-27
fix: remove pagination-related query elements on LIMIT 1

### LINKED ISSUE #1167: LIMIT argument is incorrectly set to 2 instead of 1
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/keto/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/keto/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

Calls to `checkDirect` perform a `GetRelationTuples` query with `x.WithSize(1)` to check whether a tuple was found or not.

https://github.com/ory/keto/blob/a39325ce6ad9a401731cfcf5961932232f92145e/internal/check/engine.go#L162-L179

For some reason though, we use the limit + 1 in the query:

https://github.com/ory/keto/blob/a39325ce6ad9a401731cfcf5961932232f92145e/internal/persistence/sql/relationtuples.go#L215-L218

So what should be `LIMIT 1` (we have a binary question - does it exist or not?) we end up with `LIMIT 2`

### Reproducing the bug

Look at a `GetActiveProject` trace in Grafana.

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

master

### On which operating system are you observing this issue?

None

### In which environment are you deploying?

None

### Additional Context

_No response_

aeneasr:
This patch removes query components which cause CockroachDB to perform a full-range query across all leaseholders for LIMIT 1 queries. Essentially, CockroachDB will query all potential nodes which might hold data when ORDER BY is used, or when the LIMIT indicates that there are rows missing.

This can drop the query latency to a few ms versus hundreds of ms on a CockroachDB `REGIONAL BY ROW` table, without impact on the business logic.

Closes #1167

## Further Comments

I wasn't quite sure how to write tests for this. One way would be to test the resulting SQL query, but we do not have such capabilities yet.

### DIFF
--- a/internal/persistence/sql/relationtuples.go
+++ b/internal/persistence/sql/relationtuples.go
@@ -212,10 +212,12 @@ func (p *Persister) GetRelationTuples(ctx context.Context, query *relationtuple.
 		return nil, "", err
 	}
 
-	sqlQuery := p.QueryWithNetwork(ctx).
-		Order("shard_id, nid").
-		Where("shard_id > ?", pagination.LastID).
-		Limit(pagination.PerPage + 1)
+	sqlQuery := p.QueryWithNetwork(ctx).Limit(1)
+	if pagination.PerPage != 1 {
+		sqlQuery = sqlQuery.Order("shard_id, nid").
+			Where("shard_id > ?", pagination.LastID).
+			Limit(pagination.PerPage + 1)
+	}
 
 	err = p.whereQuery(ctx, sqlQuery, query)
 	if err != nil {


aeneasr:
The tests are currently failing because they test pagination using a LIMIT 1 query. There are three options:

1. Refactor the tests to use a higher limit
2. Add a dedicated function which explicitly tests for the existence of one row
3. Add a parameter to the existing GetRelationTuples function that indicates that we do not want pagination

aeneasr:
Closing as per discussion

PR #1169, opened 2022-12-30
feat: improve tracing

alnr:
needs https://github.com/ory/x/pull/650


go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/ory/herodot v0.9.13
 	github.com/ory/jsonschema/v3 v3.0.7
 	github.com/ory/keto/proto v0.10.0-alpha.0
-	github.com/ory/x v0.0.502
+	github.com/ory/x v0.0.524-0.20221229213023-29c9e1da3e4d
```
hperl, line 26: Is the needed functionality not simply in `0.0.524`?

hperl (approved):
LGTM :tada:

internal/check/engine.go
```diff
@@ -65,7 +68,17 @@ func (e *Engine) CheckIsMember(ctx context.Context, r *relationTuple, restDepth
...
+	ctx, span := trace.SpanFromContext(ctx).TracerProvider().Tracer("keto/internal/check").Start(ctx, "Engine.CheckRelationTuple")
+	defer func() {
+		if res.Err != nil {
+			span.SetStatus(codes.Error, res.Err.Error())
+		} else {
+			span.SetAttributes(attribute.String("membership", res.Membership.String()))
+		}
+		span.End()
+	}()
+
```
aeneasr, line 81: Since the function has only two returns, why do we need a defer for this? You could also do:

```
span.SetAttributes(attribute.String("membership", res.Membership.String()))
return result

err := ...
span.SetStatus(codes.Error, res.Err.Error())
return err
```

aeneasr (changes requested):

internal/check/engine.go
alnr: There are two possible error returns from the function:
- `<-ctx.Done()`
- `<-resultCh` returns a struct with the Err field set.

Checking the named return value in the deferred call catches them both. Without that, the branch (`if res.Err != nil`) would need to move inside the `select` which is awkward.

go.mod
alnr: yep, changed.

CaptainStandby:
You changed all the license headers (probably automatically). I think this is wrong, as the year of these copyright markings normally reflect the year of first publication.
See also:
- https://stackoverflow.com/a/2391555
- https://en.wikipedia.org/wiki/Copyright_notice

We probably should fix the automatic tooling

internal/check/engine.go
aeneasr: Not sure if I understand - would this not just be:

```patch
	resultCh := make(chan checkgroup.Result)
	go e.checkIsAllowed(ctx, r, restDepth)(ctx, resultCh)
	select {
	case result := <-resultCh:
+              span.SetAttributes(attribute.String("membership", res.Membership.String()))
		return result
	case <-ctx.Done():
+               span.SetStatus(codes.Error, ctx.Err())
		return checkgroup.Result{Err: errors.WithStack(ctx.Err())}
	}
```

alnr:
> You changed all the license headers (probably automatically). I think this is wrong, as the year of these copyright markings normally reflect the year of first publication. See also:
> 
> * https://stackoverflow.com/a/2391555
> * https://en.wikipedia.org/wiki/Copyright_notice
> 
> We probably should fix the automatic tooling

I did it on purpose (with `make format`) because the format job failed in CI. You are probably correct, though.

CaptainStandby:
> I did it on purpose (with `make format`) because the format job failed in CI. You are probably correct, though.

I have created a draft for fixing this [here](https://github.com/ory/cli/pull/290)

internal/check/engine.go
alnr: in the top branch, if `result.Err` is not `nil`, we would miss that error. Not sure if that can ever happen 🤷

alnr: @aeneasr do you still want me to make this change? Otherwise this PR is ready to merge.

aeneasr: I think it's a bit clonky to read compared to the two-liner, but it does guard against future regressions. It's also just a style choice so I'm fine either way - you decide :)

PR #1203, opened 2023-01-24
chore: bump all deps

hperl:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


internal/driver/config/provider.go
```diff
@@ -167,8 +167,8 @@ func (k *Config) Set(key string, v any) error {
 func (k *Config) addressFor(endpoint EndpointType) string {
 	return fmt.Sprintf(
 		"%s:%d",
-		k.p.String("serve."+string(endpoint)+".host"),
-		k.p.Int("serve."+string(endpoint)+".port"),
+		k.p.StringF("serve."+string(endpoint)+".host", ""),
+		k.p.IntF("serve."+string(endpoint)+".port", 0),
```
zepatrik, line 171: Are these defaults defined in the schema, do we get them because of the zero-value, or why do we remove it here?

proto/go.mod
```diff
@@ -1,14 +1,24 @@
...
-go 1.16
+go 1.19
 
 require (
-	github.com/stretchr/testify v1.8.0
-	golang.org/x/net v0.0.0-20220622184535-263ec571b305 // indirect
-	golang.org/x/sys v0.0.0-20220622161953-175b2fd9d664 // indirect
-	google.golang.org/genproto v0.0.0-20220622171453-ea41d75dfa0f // indirect
-	google.golang.org/grpc v1.49.0
+	github.com/grpc-ecosystem/grpc-gateway/v2 v2.15.0
```
zepatrik, line 6: Why is this dependency added now? :thinking:

hperl: fixed, I had stale generated files in `./proto`.

internal/driver/config/provider.go
hperl: Yes, the defaults from the schema are used when the config is empty. These fallbacks would be used if there were no defaults.

Note that this is a workaround for an issue in ory/x: https://github.com/ory/x/issues/664

zepatrik (approved):

PR #1211, opened 2023-01-26
fix: tiny stuff

alnr:


internal/persistence/sql/traverser.go
```diff
@@ -145,7 +145,7 @@ func (t *Traverser) TraverseSubjectSetRewrite(ctx context.Context, start *relati
 	}
 
 	if len(relations) > 0 {
-		_, span := t.d.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.TraverseSubjectSetRewrite")
+		ctx, span := t.d.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.TraverseSubjectSetRewrite")
```
zepatrik, line 148:
```suggestion
		var span *otel.Span
		ctx, span = t.d.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.TraverseSubjectSetRewrite")
```
because of scoping?

zepatrik: Would not make a difference right now, but might later on. Or maybe just move that to the beginning of the function?

alnr: Actually you're correct! fixed, thx

zepatrik (approved):

internal/persistence/sql/traverser.go
```diff
@@ -127,6 +127,8 @@ LIMIT ?
 }
 
 func (t *Traverser) TraverseSubjectSetRewrite(ctx context.Context, start *relationtuple.RelationTuple, computedSubjectSets []string) (res []*relationtuple.TraversalResult, err error) {
+	ctx, span := t.d.Tracer(ctx).Tracer().Start(ctx, "persistence.sql.TraverseSubjectSetRewrite")
+	defer otelx.End(span, &err)
```
hperl, line 131: I get why you want to put the tracing stuff at the top, but the reason why I put it in front of the query was because sometimes the query does not even execute, and I want to have some way to count the number of DB queries that resulted from one `check` request.

If you want to put the tracing up top I think we should extract the query out into its own function that then has the tracing up top. WDYT?

hperl (changes requested):

internal/persistence/sql/traverser.go
zepatrik: You could also just add an attribute to the span once you run into that case. Would be a bit better IMO.

alnr: The DB query (if it happens) would get it's own child spawn though, wouldn't it?

hperl (approved):
Nice, I found a better way to count DB queries now, so this is perfect :+1:

PR #1219, opened 2023-01-30
feat: return bad request on DELETE body

hperl:
```
BREAKING CHANGES: This patch changes the behavior of the DELETE /relation-tuples API to reject requests with a 400 Bad Request that have:
* a request body (because no request body is consumed, so supplying a request body is interpreted as an error),
* extra query parameters (which might indicate mistyping a parameter), or
* no namespace specified (which would delete all tuples)
```

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


aeneasr (changes requested):
My recommendation would be to restrict deleting all tuples in the manager - i.e. require at least one parameter to be set.

internal/relationtuple/transact_server_test.go
```diff
@@ -218,6 +215,74 @@ func TestWriteHandlers(t *testing.T) {
...
+					http.MethodDelete,
+					ts.URL+relationtuple.WriteRouteBase+"?"+q.Encode(),
+					strings.NewReader("some body"))
+				require.NoError(t, err)
+
+				assertBadRequest(t, req)
+				assertTuplesExist(t)
+			})
+
+			t.Run("case=bad request query param misspelled", func(t *testing.T) {
```
aeneasr, line 274: Add one test where no parameter is sent?

hperl: added!

PR #1238, opened 2023-02-13
feat: gRPC gateway

hperl:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


proto/ory/keto/relation_tuples/v1alpha2/check_service.proto
```diff
@@ -17,7 +20,43 @@ option php_namespace = "Ory\\Keto\\RelationTuples\\v1alpha2";
...
+        body: "*"
+      }
+      additional_bindings: {
+        post: "/relation-tuples/check/openapi"
+        body: "*"
+      }
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      consumes: "application/x-www-form-urlencoded"
+      consumes: "application/json"
```
zepatrik, line 38: This says we can submit the payload either as a URL encoded form, or JSON for any of the endpoints defined above right?

proto/ory/keto/relation_tuples/v1alpha2/check_service.proto
```diff
@@ -35,9 +74,26 @@ message CheckRequest {
   // The relation between the Object and the Subject.
   string relation = 3 [deprecated = true];
   // The related subject in this check.
-  Subject subject = 4 [deprecated = true];
+  Subject subject = 4 [
+    deprecated = true,
+    (google.api.field_visibility).restriction = "NO_SWAGGER"
```
zepatrik, line 79: This results in the fields not included in the openapi spec?

proto/ory/keto/relation_tuples/v1alpha2/check_service.proto
```diff
@@ -35,9 +74,26 @@ message CheckRequest {
...
+  ];
+
+  oneof rest_api_subject {
```
zepatrik, line 82: I guess this is required because the tuple in gRPC is wrapped in the tuple field. Can't we use the `body: "tuple"` annotation instead? The way I understand the docs is that we can specify which field(s) are in the body (and potentially path or query). In this case we would have the `tuple` and `max-depth` fields in the body.

internal/e2e/sdk_client_test.go
```diff
@@ -286,7 +288,7 @@ func (c *sdkClient) waitUntilLive(t require.TestingT) {
 	for err != nil {
 		resp, _, err = c.getReadClient().MetadataApi.IsReady(c.requestCtx()).Execute()
 	}
-	require.Equal(t, "ok", resp.Status)
+	require.Equal(t, "SERVING", resp.Status)
```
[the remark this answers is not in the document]
hperl: Yes this is breaking, but only if you look at the content and not the code. Working around this would be hacky, as the "SERVING" response is standard in gRPC land.

proto/ory/keto/relation_tuples/v1alpha2/check_service.proto
hperl: Yes, correct.

proto/ory/keto/relation_tuples/v1alpha2/check_service.proto
hperl: Yes, because for POST we go through the body, for GET we go through the URL encoded query params.

internal/e2e/full_suit_test.go
```diff
@@ -150,8 +150,9 @@ func TestServeConfig(t *testing.T) {
 	req, err := http.NewRequest(http.MethodOptions, "http://"+reg.Config(ctx).ReadAPIListenOn()+relationtuple.ReadRouteBase, nil)
 	require.NoError(t, err)
 	req.Header.Set("Origin", "https://ory.sh")
+	req.Header.Set("Access-Control-Request-Method", http.MethodGet)
 	resp, err := http.DefaultClient.Do(req)
 	require.NoError(t, err)
-	require.Equal(t, http.StatusOK, resp.StatusCode)
+	require.Equal(t, http.StatusNoContent, resp.StatusCode)
```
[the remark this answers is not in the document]
hperl: No, this was just a broken test that has nothing to do with the gRPC gateway. Previously, the `Access-Control-Request-Method` was missing.

proto/ory/keto/relation_tuples/v1alpha2/check_service.proto
hperl: The problem is that we also have a `GET` endpoint for check, and the mapping has to handle this case as well.

With `check` we really have too many endpoints. For a next version, I would like to have only one endpoint and also define the `OpenAPI` behaviour as a flag on the request.

internal/driver/daemon.go
```diff
@@ -318,27 +313,51 @@ func (r *RegistryDefault) allHandlers() []Handler {
...
+}
+
 func (r *RegistryDefault) ReadRouter(ctx context.Context) http.Handler {
 	n := negroni.New()
 	for _, f := range r.defaultHttpMiddlewares {
 		n.UseFunc(f)
 	}
 	n.Use(reqlog.NewMiddlewareFromLogger(r.l, "read#Ory Keto").ExcludePaths(healthx.AliveCheckPath, healthx.ReadyCheckPath))
 
+	conn, err := grpc.DialContext(ctx, r.Config(ctx).ReadAPIListenOn(), grpc.WithTransportCredentials(insecure.NewCredentials()))
```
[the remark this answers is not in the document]
hperl: You are right, this is still somewhat unsound, especially in the case where the gRPC server has an additional authenticating middleware or uses TLS.

The sidecar case should be fine (something like Envoy), as the HTTP requests still get forwarded to the Keto container where the gRPC gateway runs.

I think the best way is to spin up a second local gRPC server just for the gRPC gateway requests. WDYT?

cmd/expand/root_test.go
```diff
@@ -25,7 +25,7 @@ func TestExpandCommand(t *testing.T) {
 				"--"+cmdx.FlagFormat, string(cmdx.FormatJSON),
 				"--insecure-skip-hostname-verification=true",
 			)
-			assert.Equal(t, "null\n", stdOut)
+			assert.Equal(t, "{\"type\":\"unspecified\",\"subject\":null,\"tuple\":null,\"children\":[]}\n", stdOut)
```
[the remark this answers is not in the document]
hperl: fixed.

internal/driver/daemon.go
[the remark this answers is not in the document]
hperl: fixed.

hperl:
@zepatrik this is ready for another round of reviews while I testdrive this internally.

zepatrik (approved):

hperl:
Superseeded by https://github.com/ory/keto/pull/1659

PR #1251, opened 2023-02-21
chore: bump dependencies

zepatrik:


internal/persistence/sql/migrations/migratest/migration_test.go
```diff
@@ -134,7 +134,7 @@ func TestMigrations(t *testing.T) {
 
 			t.Run("suite=uuid_migrations", func(t *testing.T) {
 				t.Run("correct types", func(t *testing.T) {
-					ctx, cancel := context.WithTimeout(ctx, 20*time.Second)
+					ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
zepatrik, line 137: Not sure why, but this kept timing out for me locally :shrug:

PR #1314, opened 2023-04-13
feat: improve emitting of events

CaptainStandby:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


internal/x/events/events.go
```diff
@@ -9,27 +9,26 @@ import (
...
-	"github.com/ory/keto/internal/x"
 )
 
-type Event string
-
 const (
-	RelationtuplesCreated Event = "RelationtuplesCreated"
-	RelationtuplesDeleted Event = "RelationtuplesDeleted"
-	RelationtuplesChanged Event = "RelationtuplesChanged"
+	EventRelationtuplesCreated semconv.Event = "RelationtuplesCreated"
```
hperl, line 15: The `Event` prefix is not needed here IMHO because the package is already named `event`, so it will be always referenced as `events.EventRelationtuplesCreated`.

internal/driver/daemon.go
```diff
@@ -323,6 +325,7 @@ func (r *RegistryDefault) ReadRouter(ctx context.Context) http.Handler {
 	for _, f := range r.defaultHttpMiddlewares {
 		n.UseFunc(f)
 	}
+	n.UseFunc(semconv.Middleware)
```
hperl, line 328: Can we just add this to the `defaultHttpMiddlewares`?

hperl (changes requested):
Looks quite good already with some minor remarks :)

internal/x/events/events.go
CaptainStandby: True. I wanted to stick to the pattern that was established in ory/x, but this seems really superfluous.

internal/driver/daemon.go
CaptainStandby: I already looked into it and had a some difficulties finding an appropriate place to do this. Do you have any recommendation where to best do it?

x/events/events.go
```diff
@@ -0,0 +1,34 @@
...
+	RelationtuplesCreated semconv.Event = "RelationtuplesCreated"
+	RelationtuplesDeleted semconv.Event = "RelationtuplesDeleted"
+	RelationtuplesChanged semconv.Event = "RelationtuplesChanged"
+
+	PermissionsExpanded semconv.Event = "PermissionsExpanded"
+	PermissionsChecked  semconv.Event = "PermissionsChecked"
+)
+
+// Emit adds an event to the current span in the context.
+func Emit(ctx context.Context, event semconv.Event, opt ...attribute.KeyValue) {
```
alnr, line 24: See https://github.com/ory/kratos/pull/3235#discussion_r1166526262

alnr (changes requested):

internal/driver/daemon.go
CaptainStandby: I'll leave this as is to get this thing merged.

alnr (approved):

aeneasr (approved):

PR #1340, opened 2023-06-13
feat: write to UUID mapper and relation tuples in one SQL transaction

alnr:
This wraps an SQL transaction around the UUID mapper's and the relation tuple manager's write operations.

Previously, we would insert into the UUID mapping table outside the SQL transaction where we updated/inserted the relation tuples. This is now one transaction.

Beside the obvious benefit of the whole operation being atomic, this improves performance in CockroachDB global tables, where each write operation outside of a transaction takes at least the time to commit the implicit transaction (100++ ms). Inside a transaction, inserts are faster and we have to pay the commit latency only once.

I'm open suggestions on how to better cut across the abstraction layers (handler vs manager/persister) 🤷 .

Before:

image

After:
image

Note: in these traces, the `COMMIT`-equivalent of CRDB is the `sql-conn-exec` before the `sql-tx-commit` span.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,14 +3,12 @@ linters:
   enable:
     - gosec
     - goimports
-    - deadcode
     - errcheck
     - gosimple
     - ineffassign
     - staticcheck
     - typecheck
     - unused
-    - varcheck
 
 linters-settings:
   goimports:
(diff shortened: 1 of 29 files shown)


hperl (approved):
LGTM! I really like how you implemented this :).

So if I understood correctly, every time we want to do things e.g. from a handler, we now request a `Transactor`, which wraps the closure in a transaction.

alnr:
> LGTM! I really like how you implemented this :).

Thanks ;)


> So if I understood correctly, every time we want to do things e.g. from a handler, we now request a `Transactor`, which wraps the closure in a transaction.

Yep, pretty much. For now, I've changed only`TransactTuples`. Maybe there are more instances where this approach makes sense?

zepatrik (approved):
Very nice :+1:

PR #1393, opened 2023-08-04
feat: clearer error messages when not using block

bittermandel:
When using `keto status` to check server health, it is not clear why the server might not be healthy. For example, if accessing the server through an insecure connection, the command would log `NOT_SERVING` and return 0, rather than `Unable to get a check response: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: EOF` with exit status 255.

This PR improves this slightly by logging the actual error that the gRPC client returns rather than wrapping it in a NOT_SERVING error. It also makes sure that the status command respects block and doesn't add a timeout by default, as otherwise the error will always be "Context deadline exceeded".

I am very happy to chat about how I can add sufficient docs and tests for these changes!

PS: The quality of this PR can most likely be improved and I'm happy to take feedback on how to do that.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.
- [x] I have added or changed [the documentation](https://github.com/ory/docs).


hperl (approved):
LGTM! Thank you for your contribution!

PR #1429, opened 2023-09-15
feat: cache OPL when loading from HTTP(S)

alnr:
https://github.com/ory-corp/cloud/issues/5319

### DIFF
--- a/internal/driver/config/opl_cache_test.go
+++ b/internal/driver/config/opl_cache_test.go
@@ -1,0 +1,52 @@
+// Copyright © 2023 Ory Corp
+// SPDX-License-Identifier: Apache-2.0
+
+package config
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/ory/x/configx"
+	"github.com/ory/x/logrusx"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewOPLConfigWatcher(t *testing.T) {
+	hits := 0
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		hits++
+		io.WriteString(w, testOPL)
+	}))
+	t.Cleanup(ts.Close)
+	ctx := context.Background()
+	cfg, err := NewDefault(ctx, nil, logrusx.New("", ""), configx.SkipValidation())
+	require.NoError(t, err)
+	cw, err := newOPLConfigWatcher(ctx, cfg, ts.URL)
+	require.NoError(t, err)
+	require.Equal(t, 1, hits, "HTTP request made")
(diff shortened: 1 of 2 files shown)


internal/driver/config/opl_cache_test.go
```diff
@@ -0,0 +1,53 @@
...
+	require.NoError(t, err)
+	cw, err := newOPLConfigWatcher(ctx, cfg, ts.URL)
+	require.NoError(t, err)
+	require.Equal(t, 1, hits, "HTTP request made")
+	_, err = cw.GetNamespaceByName(ctx, "User")
+	require.NoError(t, err)
+	_, err = cw.GetNamespaceByName(ctx, "Document")
+	require.NoError(t, err)
+
+	time.Sleep(1 * time.Second) // warm cache
```
aeneasr, line 37: Please use `cache.Wait()` instead, otherwise this will be flaky

aeneasr (changes requested):

internal/driver/config/opl_cache_test.go
alnr: Uh, brilliant!

internal/driver/config/opl_config_namespace_watcher.go
```diff
@@ -33,7 +36,14 @@ type (
...
 )
 
-var _ namespace.Manager = (*oplConfigWatcher)(nil)
+var (
+	_        namespace.Manager = (*oplConfigWatcher)(nil)
+	cache, _                   = ristretto.NewCache(&ristretto.Config{
+		MaxCost:     20_000_000, // 20 MB max size, each item ca. 10 KB => max 2000 items
+		NumCounters: 20_000,     // max 2000 items => 20000 counters
+		BufferItems: 64,
+	})
```
aeneasr, line 45: Ristretto config is a science for itself. I recommend this:

```suggestion
	cache, _                   = ristretto.NewCache(&ristretto.Config{
		NumCounters:        maxItems * 10, // Counters should be 10x max cost (see ristretto docs).
		MaxCost:            maxItems,
		BufferItems:        64,
		Metrics:            true, // consider false
		IgnoreInternalCost: true,
		Cost: func(value interface{}) int64 {
			return 1
		},
 	})
``` 

And then set `maxItems` to `20MB / 10kb = 2048`. Although I'd argue that up to 500MB is fine for this cache.

zepatrik (approved):

internal/driver/config/opl_config_namespace_watcher.go
alnr: Is this better, though? OPL definitions are not all the same size, yet their individual cost is easy to calculate.

LMK.

aeneasr: ok, we can also fine-tune this by checking how this is doing in our tracing

Issue #812, opened 2022-01-10
Batch check request


zepatrik:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/keto/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/keto/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

In some cases it makes sense to do multiple checks at the same time. This can be used e.g. to create a list of all permissions for the UI, or similar things.

### Describe your ideal solution

We can provide a batch check API. It would make sense especially for cases where subjects and/or objects are the same, so that we can reuse already done queries internally.

### Workarounds or alternatives

Doing multiple check requests. This has the overhead of the transport, as well as reduced performance optimization possibilities.

### Version

all

Fylipp:
This would help me and my team with the product we are building. We have paginated endpoints that return lists of resources that are sorted and filtered via Mongo queries. Afterwards, we would like to check the user permissions of all items in the page, but at the moment we are forced to do a separate check per item. Implementing this would allow us to only call the API once and speed things up.

ryukinix:
That feature would help my team. We have a feature-store and complex data system, we are in the process of making this a platform beyond the company internal usage for data science, but a product (external usage). At some point we have to list a catalog of datasets that the user have access. We have a 1000+ datasets.

I thought initially that we could do something similar with keto using `GET /relation-tuples/expand` or `GET /relation-tuples` by querying with a fixed `namespace` and `subject_id`, but for my surprise it didn't what I expected.

istarkov:
Or the ability to `expand` without setting relation would help too. The case that UI depends on the current permit and in our case it causes 8 checks. i.e. does the user have `xxx` permit, does the token have `xxx` permit, where `xxx` =  list[vew, edit, build, own]

shashimalcse:
We are encountering a similar issue where we need to send multiple check requests to keto. It would be more efficient to have a batch request in order to reduce the transport overhead.

[...]

PR #1521, opened 2024-04-15
Batch check relations

patrickduffy95:
## Related issue(s)
https://github.com/ory/keto/issues/812

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments
NOTE: This is a "draft" PR that isn't complete and is meant to get early feedback on the proposed set of changes.

This change adds REST and gRPC endpoints for batch checking relations. The endpoint accepts a list of relation tuples to check, iterates through them, and returns a list of allowed responses.

**REST API**
New endpoint:
`POST /relation-tuples/batchCheck?max-depth=<depth>`
Request body:
```
{
    "tuples": [
        {
            "namespace": <namespace>,
            "object": <object>,
            "relation": <relation>,
            "subject_id": <subject_id>,
            "subject_set": <subject_set>
        }
    ]
}
```
Response:
```
{
    "results": [
        {
            "allowed": true
        }
    ]
}
```

**gRPC**
New RPC:
`CheckService/BatchCheck`
Request:
```
// The request for a CheckService.BatchCheck RPC.
// Checks a batch of relations.
message BatchCheckRequest {
  repeated RelationTuple tuples = 1;

  // This field is not implemented yet and has no effect.
  bool latest = 2;
  // This field is not implemented yet and has no effect.
  string snaptoken = 3;
  int32 max_depth = 4;
}
```
Response
```
// The response for a CheckService.Check rpc.
message BatchCheckResponse {
  //
  repeated CheckResponse results = 1;
}
```

Questions I'd like feedback on:
- Should we limit the number of tuples this batch API accepts?
- Should we parallelize the check requests?
  - A more optimized implementation that reduces the number of calls to the DB rather than iterating through the request tuples is probably more than I'd want to tackle
- Is implementing only the POST okay for the REST API?


patrickduffy95:
@alnr would you be able to let me know if this PR is on the right track?

aeneasr:
Hello, sorry for kot responding here. I think this feature is grand! @zepatrik and @hperl are finishing up some work on our end and can probably check out this PR in a week or two.

PrimeDominus:
Thanks for implementing this. We've hit a roadblock and without this feature we have to write a bunch of custom code in our server which is getting cumbersome to maintain. Really looking forward to trying this.

patrickduffy95:
> Hello, sorry for kot responding here. I think this feature is grand! @zepatrik and @hperl are finishing up some work on our end and can probably check out this PR in a week or two.

Thanks @aeneasr! Do you have a better idea now of when they might be able to take a look at it?

internal/check/handler.go
```diff
@@ -329,3 +331,121 @@ func (h *Handler) Check(ctx context.Context, req *rts.CheckRequest) (*rts.CheckR
...
+			Allowed: allowed,
+		}
+	}
+
+	return results, nil
+}
+
+func (h *Handler) BatchCheck(ctx context.Context, req *rts.BatchCheckRequest) (*rts.BatchCheckResponse, error) {
+	responses := make([]*rts.CheckResponse, len(req.Tuples))
+	for i, reqTuple := range req.Tuples {
```
hperl, line 432: The logic is duplicated from the REST handler here. Ideally we'd implement this once and have gRPC and REST call that function.

hperl (changes requested):
Thank you very much for your contribution! This looks very good already.

Besides the comments below, the main thing that is missing are some tests. I'd like to see tests for the happy path, going over the batch limit (bad request), graceful handling of on errornous tuple in the batch (should not fail the whole batch, but just that check), empty batch, and whatever edge case you can think of.

I left some comments below. As for your questions, here's my take on it:

> Should we limit the number of tuples this batch API accepts?

Yes, please make it configurable with a conservative default (10?)

> Should we parallelize the check requests?

Yes, I'd make the number of concurrent checks another parameter.

> Is implementing only the POST okay for the REST API?

Yes, that will suffice.

patrickduffy95:
Thanks a lot for the review @hperl. I will let you know once this PR is updated.

internal/check/handler.go
```diff
@@ -329,3 +331,121 @@ func (h *Handler) Check(ctx context.Context, req *rts.CheckRequest) (*rts.CheckR
...
+			results[i] = &CheckPermissionResult{
+				Allowed: false,
+			}
+			continue
+		} else if err != nil {
+			return nil, err
+		}
+		allowed, err := h.d.PermissionEngine().CheckIsMember(ctx, t[0], maxDepth)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
patrickduffy95: Adding more context to the error is a good idea. The question then is whether to return errors at the tuple level vs for the API as a whole. If there are situations where clients might expect an error and would want to treat an error the same as receiving a `false` allowed value, I could see a benefit in continuing on error and returning for each tuple if an error was generated. As far as I know though, that isn't a normal case, so I will just add tuple details to the error message.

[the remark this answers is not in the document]
patrickduffy95: Changed my mind and went for proceeding with checking the whole batch and adding the error to the response for each tuple

cmd/status/root_test.go
```diff
@@ -37,7 +37,7 @@ func TestStatusCmd(t *testing.T) {
 				defer cancel()
 
 				stdErr := cmdx.ExecExpectedErrCtx(ctx, t, newStatusCmd(), "--"+FlagEndpoint, string(serverType), "--"+ts.FlagRemote, ts.Addr[:len(ts.Addr)-1])
-				assert.Equal(t, "context deadline exceeded", stdErr)
+				assert.Contains(t, stdErr, "context deadline exceeded")
```
patrickduffy95, line 40: I think this started failing b/c of the dependency bump

patrickduffy95:
@hperl I have update the PR based on your feedback and it is now ready for review.

PrimeDominus:
I'm really looking forward to this functionality being one Ory Network. @hperl is there anything else that needs to be done before this can be merged?

internal/check/handler.go
```diff
@@ -329,3 +354,155 @@ func (h *Handler) Check(ctx context.Context, req *rts.CheckRequest) (*rts.CheckR
...
+}
+
+// doBatchCheck is the HTTP entry point for checking batches of tuples
+func (h *Handler) doBatchCheck(ctx context.Context, body io.Reader, query url.Values) ([]*CheckPermissionResultWithError, error) {
+	maxDepth, err := x.GetMaxDepthFromQuery(query)
+	if err != nil {
+		return nil, err
+	}
+	parallelizationFactor := defaultBatchCheckParallelizationFactor
+	if query.Get(parallelizationFactorQueryParam) != "" {
```
hperl, line 431: Sorry for the misunderstanding, but I meant having this as a config parameter in the config (where you also added the max batch size). This is more a parameter the server admin sets to cap the max load of Keto than for a client to set, in my mind.

internal/check/engine.go
```diff
@@ -264,3 +267,46 @@ func (e *Engine) astRelationFor(ctx context.Context, r *relationTuple) (*ast.Rel
...
+		return nil, errors.New("invalid parallelization factor")
+	}
+
+	wg := &sync.WaitGroup{}
+	sem := semaphore.NewWeighted(int64(parallelizationFactor))
+
+	mapper := e.d.ReadOnlyMapper()
+	results := make([]checkgroup.Result, len(tuples))
+	for outerIndex, outerTuple := range tuples {
+		sem.Acquire(context.Background(), 1) // Pass in background context to guarantee this won't return an error
```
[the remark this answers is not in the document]
patrickduffy95: Nice catch on the race condition. I've updated it to use `sync.errgroup.Group`

internal/check/handler.go
patrickduffy95: No worries, I've updated it to be a config parameter

hperl (approved):
This looks perfect now! Thanks for all the good work! 🎉

rbnbr:
Hey, thank you for the great feature!

Now with this being merged in the master branch, could we also extend the Rest HTTP API accordingly (https://www.ory.sh/docs/keto/reference/rest-api) and also make it available to the keto-client(s), especially for go (https://github.com/ory/keto-client-go)?

Thank you!

ryukinix:
Very interesting! We were waiting for this since 2022. Awesome work!

jcottongit:
@hperl Cheers for getting this in, can we get a docker image released with this feature included?

PR #1525, opened 2024-04-29
fix: missing persister when initializing without network

alnr:


go.mod
```diff
@@ -4,7 +4,7 @@ module github.com/ory/keto
 
 replace github.com/ory/keto/proto => ./proto
 
-replace github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.12
+replace github.com/opencontainers/runc => github.com/opencontainers/runc v1.2.0-rc.1
```
hperl, line 7: IIRC this is not compatible with Go 1.22 :(

hperl (changes requested):

go.mod
alnr: Wait, but 1.1.12 is fine?

hperl: Yes, that's fine, but that has a CVE associated to it. But we can't upgrade, because 1.2.0-rc.1 is not compatible with Go 1.22.

https://github.com/opencontainers/runc/releases/tag/v1.2.0-rc.1

> NOTE: runc currently will not work properly when compiled with Go 1.22 or
> newer. This is due to some unfortunate glibc behaviour that Go 1.22
> exacerbates in a way that results in containers not being able to start on
> some systems. https://github.com/opencontainers/runc/issues/4233

alnr: done

hperl (approved):

PR #1659, opened 2024-12-18
feat: unified APIs with REST, gRPC and Connect RPC

hperl:
This PR unifies the API service implementations between REST and gRPC, and adds with the support for Connect RPC a third protocol, all via [Vanguard](https://github.com/connectrpc/vanguard-go).

The advantages are as follows:
- Unified, fully typed service implementations powered by gRPC. No more manual request parsing, no more duplicated code between the REST and gRPC implementations.
- Unified middleware stack: We can use the Oathkeeper gRPC middleware to protect all API requests.
- Support for Connect RPC, which offers modern SDKs for languages that the OpenAPI client struggles with (Kotlin, Swift).
- Protobuf as the IDL, no more go-swagger

Requests flow through the system as follows:

1. All services land on a handler that first checks if tthe path has been registered directly, for example for the Ory-wide health and version paths, or other paths that are REST specific. If so, the request is forwarded to that route's handler. This is always an escape hatch for REST APIs to remain backwards compatibility.
2. Next, for REST and Connect RPC, the request is transcoded into gRPC.
3. Next, the gRPC service implementation is called.
4. Finally, for REST, a middleware converts the gRPC error (if any) into a Herodot error, and optionally sets the response status, and location header to one specified by the service implementation

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.


aeneasr:
The API specs all remain the same?

internal/x/dbx/dsn_testutils.go
```diff
@@ -106,47 +105,47 @@ func createDB(t testing.TB, url string, dbName string) (err error) {
...
-		if cockroach != "" {
-			dsns = append(dsns, &DsnT{
-				Name:        "cockroach",
-				Conn:        cockroach,
-				MigrateUp:   true,
-				MigrateDown: true,
-			})
-		}
-	}
+	//if !testing.Short() {
```
aeneasr, line 108: Revert?

spec/api.json
```diff
@@ -473,40 +469,31 @@
...
       },
       "patch": {
-        "description": "Use this endpoint to patch one or more relationships.",
         "operationId": "patchRelationships",
         "requestBody": {
           "content": {
             "application/json": {
               "schema": {
                 "items": {
-                  "$ref": "#/components/schemas/relationshipPatch"
```
aeneasr, line 499: Breaking change

spec/api.json
```diff
@@ -517,33 +504,12 @@
               }
             },
             "description": "errorGeneric"
-          },
-          "404": {
```
aeneasr, line 521: Missing?

aeneasr (commented):
Seems like there are a few breaking changes in the OpenAPI spec, but not too many

spec/api.json
hperl: fixed!

